PR #4, opened 2015-08-29
using JERKY interface (websocket) instead of JERK (socket)

yazgoo:


fommil:
@yazgoo just merge these yourself :-)

PR #5, opened 2015-08-29
adding unit tests - drone.io support

yazgoo:


fommil:
We can add all the ruby setup stuff indie the ENSIME docker image if you'd like. That'll really speed up the tests.

yazgoo:
Sure,
Here is the commmand under drone.io to setup the required gems:

bundle install

fommil:
docker image is in `ensime-docker`. Send a PR to the `ensime-1.x` branch and it'll show up like magic after an hour or so in CI.

PR #7, opened 2015-08-30
drone io fix

yazgoo:


yazgoo:
Creating vim plugin and ruby gem - not yet validated with drone

PR #37, opened 2015-09-16
Add gradle instructions

xanderdunn:
This makes it clear that gradle can be used instead of sbt

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ You should also export your BROWSER variable, for example in your bashrc:
 
 All the following commands should be ran from your scala directory.
 
-First you need ensime sbt plugin:    
+First you need ensime sbt plugin (Or if you're using gradle, you can use [ensime-gradle](https://github.com/ensime/ensime-gradle)):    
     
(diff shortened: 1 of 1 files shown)


yazgoo:
Does this account for the fact that sbt is used to generate ensime classpath ?

xanderdunn:
Can you show me an example .ensime of what this looks like?  

[Here](https://gist.github.com/xanderdunn/3755eaaabac62400c018) is the .ensime that the gradle plugin produced for my project xanderdunn/reinforcementlearning.

yazgoo:
The part using sbt is not the one doing .ensime generation, which should be done by the user (for now).
The problem is with launching ensime server, which requires to find out ensime classpath, which is done with sbt:
https://github.com/ensime/ensime-vim/blob/768892b009ff753b63baf8173ba33de1838b59bf/ensime_launcher/__init__.py#L68

xanderdunn:
Realizing now that sbt is an integral part of the plugin's functionality, this is not sufficient to provide gradle support.  Perhaps a larger pull request in the future can handle this functionality with gradle, but for now sbt will do.

PR #49, opened 2015-09-18
import ensime_launcher from local

todesking:
Dependency: https://github.com/ensime/ensime-vim/pull/48


yazgoo:
This is such a good idea !

PR #57, opened 2015-09-30
do not start ensime server is it is already running externally

kstep:
I run ensime server manually for my projects with systemd user unit, so I don't need to (re)start it every time I restart vim.

This patch allows me to re-use already existing server.

### DIFF
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -30,8 +30,12 @@ function! ensime#with_current_client(proc) abort
     return s:call_plugin('with_current_client', [a:proc])
 endfunction
 
-function! ensime#fun_en_complete_func(args) abort
-    return s:call_plugin('fun_en_complete_func', [a:args])
+function! ensime#is_scala_file() abort
+    return s:call_plugin('is_scala_file', [])
+endfunction
+
+function! ensime#fun_en_complete_func(findstart, base) abort
+    return s:call_plugin('fun_en_complete_func', [a:findstart, a:base])
 endfunction
 
 function! ensime#au_vimleave(filename) abort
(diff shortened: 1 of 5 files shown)


kstep:
The second commit fixes an error for me: when I use `:EnDeclaration` new buffer with Scala file is opened OK, but its filetype is unset, so I don't get syntax highlighting etc.

jvican:
Cool, @kstep, I was about to report the bug of `:EnDeclaration`.

yazgoo:
Thanks,
On vim:
I'm getting this error when running EnDocUri

````
E605: Exception not caught: <type 'exceptions.TypeError'>:cannot concatenate 'str' and 'int' objects
````

I'm also getting `Using list as number` when trying autocomplete.

kstep:
Could you give me a choice snippet you use ensime-vim on to get such results? Thank you.

yazgoo:
I'm using this project:
https://github.com/pocorall/scaloid/ running:

vim src/scaloid/example/HelloScaloid.scala

yazgoo:
Again thanks for your work :+1: 
Looks like you're editing autoload file which is a generated file, via https://github.com/yazgoo/neo2vim/pulls see developer howto https://github.com/ensime/ensime-vim/blob/master/README.md

kstep:
Thanks for pointers! I cleaned up things a little, updated file in rpython and regenerated autoload part.
It should look better now.

kstep:
To sum things up, this PR now has the changed:

- re-use existing ensime server if it's already listening,
- EnDeclaration filetype setup fix
- autocompletion fixes

PR #60, opened 2015-10-01
Fix completion

todesking:


autoload/ensime.vim.py
```diff
@@ -280,11 +279,11 @@ def autocmd_handler(self, filename):
...
     # @neovim.function('EnCompleteFunc', sync=True)
-    def complete_func(self, args):
-        if args[0] == '1':
+    def complete_func(self, findstart, base):
+        if findstart == '1':
             self.complete()
             line = self.vim.eval("getline('.')")
-            start = self.cursor()[1] - 1
+            start = self.cursor()[1]
             pattern = re.compile('\a')
```
kstep: This works by accident only, as there are almost never matching against `chr(0x07)` happens.
It should be something like `r'\b'`, like in #57 (90eb22874fcdf9f6f3e514068a90bc012cc91e83).

yazgoo:
Thanks,

When I run under vim, though:

S`ctrl+X``ctlr+U`

The completion is put after the first character.
For example if I choose SView on the list, I get

SSview

kstep:
Please see #57, that version works right for me.
As I pointed out in the line comment above, the code to find the start of word to complete is incorrect here.

PR #95, opened 2015-10-19
Fix ReadMe install instructions

JD557:
There was a small typo in the `echo` command of the Read Me instructions, which made the command impossible to run.

This addresses that issue.


yazgoo:
Thanks

PR #107, opened 2015-11-06
Additional troubleshooting tips for Vim users

arjunsw:
This will help pure vim users when they stumble across "initializing" message due to incomplete ensime-vim installation. 
In my case, I saw an error in generating sbt file by the ensime-vim plugin. Adding blank lines between the sbt statements in the build_sbt method inside ensime_launcher/__init__.py lead to invalid SBT file.  

Thereafter SBT started installing everything (note: I have SBT 0.13.9 and wondered why this matters).

While installing dependencies, some Netbeans modules were not found so added additional settings in my build.sbt and the whole process completed smoothly.

BTW, I just added to the README. Would like to change the the above py file directly if it is ok in future commits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ ENSIME for the Editor of the Beast (Vim)
 # howto
 
 You need `websocket-client` python package:
-
+    
     $ sudo pip install websocket-client
 
 You should also export your BROWSER variable, for example in your bashrc:
@@ -88,7 +88,7 @@ You need not use `Vundle`. Just go to `~/.vim/bundle` and clone `ensime-vim`.
 
 Inside `~/.vim/bundle/ensime-vim` folder, edit the following file:
 
-  vim ensime_launcher/__init__.py 
+    vim ensime_launcher/__init__.py 
 
 Find the method `build_sbt`. In the `src` string, add blank lines between
 `SBT` statements (though it should not matter for `SBT 0.13.7` onwards, there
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,6 +82,33 @@ And export them to vim plugin format via:
...
+# Additional helpful tips for Vim users
+
+You need not use `Vundle`. Just go to `~/.vim/bundle` and clone `ensime-vim`.
+
+Inside `~/.vim/bundle/ensime-vim` folder, edit the following file:
+
+    vim ensime_launcher/__init__.py 
+
+Find the method `build_sbt`. In the `src` string, add blank lines between
+`SBT` statements (though it should not matter for `SBT 0.13.7` onwards, there
```
yazgoo: I think this should be fixed directly in build_sbt() instead, shouldn't it ?

README.md
```diff
@@ -80,6 +82,33 @@ And export them to vim plugin format via:
 
 All merges should be done on dev branch before being merged onto master
 
+# Additional helpful tips for Vim users
+
+You need not use `Vundle`. Just go to `~/.vim/bundle` and clone `ensime-vim`.
+
```
yazgoo: This is just what vundle does, why do it manually ?

yazgoo:
Maybe we could try and merge it with the howto section, what do you think ?

README.md
arjunsw: agreed

README.md
arjunsw: Just a thought for people who already have Pathogen. Could be a note for people who already have Pathogen.

arjunsw:
We should just patch the code. That way the current README stays and things work seamlessly with new code. BTW, instead of build.sbt having the netbeans module dependencies, we could put that too in the vim plugin sbt template. I have to check that though.

yazgoo:
@arjunsw, does 62026d2 adds blank lines where you'd expect ?

yazgoo:
Closing this since documentation was updated, please re-open another issue/PR if you think there is still work to be done.

PR #118, opened 2015-11-25
Support omnifunc to be invoked with findstart as a number

ktonga:
Some times the `omnifunc` is invoked with `findstart` arg as a number `1`. An example of such kind of invocation is YCM semantic completion.

Comparing the argument against string value `'1'` causes to return a list as the result for the first invocation when it has to be a number (cursor position). So it generates an error in the caller, for instance YCM halts for a while (several seconds) and then it shows the following error in the status line:

    Omnifunc returned bad value to YCM! int() argument must be a string or a number, not 'list'

Wrapping `findstart` with `str()` is safer in the case it is a number or a string.


yazgoo:
Thanks for your contribution!

PR #124, opened 2015-12-07
fixing goto declaration and temp fix for syntax highlighting

vito-c:
- Since you have the byte offset let's use that instead [:help goto](http://vimhelp.appspot.com/motion.txt.html#%3Agoto) 
- There was also an issue with the syntax becoming disabled. I didn't have time to track down exactly where that was happening. I just re-enabled it for the time being. 
- logging via echo should be disabled it can cause the user to have to hit enter to proceed

### DIFF
--- a/rplugin/python/ensime.py
+++ b/rplugin/python/ensime.py
@@ -161,16 +161,8 @@ class EnsimeClient(object):
         if declPos["typehint"] == "LineSourcePosition":
             self.set_cursor(declPos['line'], 0)
         else: # OffsetSourcePosition
-            f = open(self.path())
-            i = 0
-            row = 0
             point = declPos["offset"]
-            for line in f:
-                i += len(line)
-                if point < i:
-                    self.set_cursor(row, point - i - len(line))
-                    break
-                row += 1
+            self.vim.command("goto %s"% (point+1))
     def get_position(self, row, col):
         result = col -1
         f = open(self.path())
@@ -323,12 +315,12 @@ class EnsimeClient(object):
         typehint = payload["typehint"]
         if typehint == "SymbolInfo":
             try:
-                self.message(payload["declPos"]["file"])
+                #self.message(payload["declPos"]["file"])
                 if self.open_definition:
                     self.clean_errors()
-                    self.vim.command(":{} {}".format("split" if self.split else "e", payload["declPos"]["file"]))
+                    self.vim.command("{} {}".format("split" if self.split else "e", payload["declPos"]["file"]))
                     self.set_position(payload["declPos"])
-                    self.vim.command("filetype detect")
+                    self.vim.command("syntax enable")
             except KeyError:
                 self.message("symbol not found")
         elif typehint == "IndexerReadyEvent":


yazgoo:
Thanks,

The color thing is not working on my side.
Otherwise, looks good to me.

vito-c:
Is the syntax highlighting showing up after you jump to declaration? Are you using vim or neovim?

Vito Cutten from mobile device

yazgoo:
Im using vim. The syntax highlighting does not show up, but after a `:e` it does. I tried adding a vim.command("e") after opening the file, but it does nothing though...
Here is my vimrc:

````vimrc
set nocompatible              " be iMproved, required
filetype off                  " required

let g:async = {'implementation' : 'c_executable'}
set rtp+=~/.vim/bundle/Vundle.vim
call vundle#begin()

Plugin 'VundleVim/Vundle.vim'

Plugin 'tpope/vim-fugitive'
Plugin 'vim-ruby/vim-ruby'
Plugin 'elzr/vim-json'
Plugin 'Valloric/YouCompleteMe'
Plugin 'ensime/ensime-vim'


call vundle#end()            " required
filetype plugin indent on    " required

syntax on
set smartindent
set expandtab
set tabstop=2
set shiftwidth=2
set cindent
if has("autocmd")
        filetype indent on
endif
filetype on
au BufNewFile,BufRead *.md set filetype=markdown
source ~/.vimrc.bepo
set rtp+=$HOME/.local/lib/python2.7/site-packages/powerline/bindings/vim/
set laststatus=2
set t_Co=256
colorscheme desert
imap ,, <Esc>

highlight Unsecable ctermbg=red guibg=red
syntax on
autocmd BufNewFile,BufRead *.scala   set ft=scala " Set syntax highlighting for .scala files
autocmd BufNewFile,BufRead *.sc      set ft=scala " Set syntax highlighting for scala worksheet files
set ruler
```

vito-c:
if you want auto syntax highlighting and filetype identification you should https://github.com/vito-c/pirate-vim/blob/master/rc/vimrc#L15 and `syntax on`

rplugin/python/ensime.py
```diff
@@ -323,12 +315,12 @@ def handle_payload(self, payload):
...
             try:
-                self.message(payload["declPos"]["file"])
+                #self.message(payload["declPos"]["file"])
                 if self.open_definition:
                     self.clean_errors()
-                    self.vim.command(":{} {}".format("split" if self.split else "e", payload["declPos"]["file"]))
+                    self.vim.command("{} {}".format("split" if self.split else "e", payload["declPos"]["file"]))
                     self.set_position(payload["declPos"])
-                    self.vim.command("filetype detect")
+                    self.vim.command("syntax enable")
```
vito-c: We shouldn't have to force syntax enable here. We need to figure out why syntax highlighting becomes disabled.

yazgoo:
removing:
````
autocmd BufNewFile,BufRead *.scala   set ft=scala " Set syntax highlighting for .scala files
autocmd BufNewFile,BufRead *.sc      set ft=scala " Set syntax highlighting for scala worksheet files
`````
Should fix it since my .vimrc already contains
````
filetype plugin indent on
"...
syntax on
```
However, it does not.
And I think I know what's going on, the problem is that an autocmd cannot use another autocmd without nested:
http://vimdoc.sourceforge.net/htmldoc/autocmd.html#autocmd-nested

yazgoo:
Using 
````
Plugin 'derekwyatt/vim-scala'
````
with
````
self.vim.command("filetype detect")
````
Works though.

yazgoo:
merged #125 with your changes for vim and `filetype detect` until we find a way nod to need it.

vito-c:
Maybe the difference is that I'm using vim plug? which requires less filetype switching? I'll test again tonight with a minimal vimrc. What is the result of `:filetype` when you are running vim? mine is `filetype detection:ON  plugin:ON  indent:ON`

yazgoo:
`filetype detection:ON  plugin:ON  indent:ON`

PR #127, opened 2015-12-09
adding type signatures to omnicomp

yazgoo:


vito-c:
do you have a screen cap? I still only see 
thing.get
`getSomeStuff [ID]`
I'd expect to see `getSomeStuff   Seq[Stuff]` or something to that effect right? Does ensime-server return that info?

vito-c:
I had this work once today but only once. I think there is something I need to do to force ensime-server to update?

yazgoo:
Maybe running `gen-ensime` would help ?

yazgoo:
capture d ecran 2015-12-11 a 13 41 03

vito-c:
how often do I need to run gen-ensime? after every code change?

rorygraves:
@vito-c  Only when you change the sbt config (e.g. add a dependency)

vito-c:
@rorygraves what about when I change a version of a dependency? 
````
  private val someVersion = "12.0.1"
  def someClient(configurations: Iterable[Configuration] = Seq(Compile)): ModuleID =
    "com.foo.bar" %% "someclient" % someVersion % concat(configurations)
````
changes from 12.0.1 to 13.0.1

rorygraves:
Yep, if you look in the .ensime file it resolves the dependencies to entires in your .ivy cache - it doesn't get dynamically updated.

vito-c:
sounds like sbt gen-ensime is something I should add to my git hooks... are you familiar with what the `EnClasspath` does? Do I need to run this everytime after I do sbt gen-ensime? I was also wondering if there is a way to send a request to the ensime server on the commandline so that I could try and debug issues faster.

rorygraves:
I don't know enough about the vim plugin to comment on EnClasspath.
the basic rule is that if you update your project structure or dependencies (update or change) you would need to rerun gen-ensime and also restart the ensime-server (it reads the .ensime file on startup).  How that interacts with the vim plugin lifecycle I'm not sure.

yazgoo:
Thanks @rorygraves ,

Here is EnClasspath description:

https://github.com/ensime/ensime-vim/blob/89cc83cd41b4a18b40dc7e45b84d6274387494f6/doc/ensime-vim.txt#L18

Basically, it:

- generates classpath for ensime server in ~/.config/classpath_project_ensime
- starts up ensime server if it is not already started.

Which means if you want to "update" ensime server, you should destroy ~/.config/classpath_project_ensime and then run EnClasspath.

yazgoo:
Merging this.

PR #130, opened 2015-12-13
fixing browser open

vito-c:

Issue #128, opened 2015-12-11
Error detected while processing CursorHold Auto commands for "*.scala" in neovim


ajkaanbal:
Hi.

The error occurs after using `:EnDocBrowse`. The documentation opens correctly in the browser but nvim shows up the error message every time I tried to move the cursor, I have to reopen neovim. 

Error message:

```
Error detected while processing CursorHold Auto commands for "*.scala":
Invalid channel "1"
```

versions:
`$ nvim --version`
NVIM 0.1.1 (compiled Dec  9 2015 13:03:38)
`$ pip freeze | grep neovim`
neovim==0.0.38

init.vim

```
call plug#begin('~/.vim/bundle')
Plug 'ensime/ensime-vim'
call plug#end()
```

ajkaanbal:
I found out what's happen

I set BROWSER=google-chrome when I change to BROWSER=firefox  it works, no more error message.

I modify this line:

https://github.com/ensime/ensime-vim/blob/master/rplugin/python/ensime.py#L303

to this: `subprocess.Popen([browser, url], stderr=subrocess.PIPE, stdout=subprocess.PIPE)`
 
And now is working also with google-chrome.

vito-c:
https://github.com/ensime/ensime-vim/pull/130

yazgoo:
Thanks @vito-c ,
Closing this, please re-open in case of problem.

PR #137, opened 2015-12-23
fix the fix: just a typo, related to #128 and #130

ajkaanbal:


vito-c:
LGTM

vito-c:
@yazgoo

yazgoo:
Thanks @vito-c and @ajkaanbal !

PR #144, opened 2016-01-06
EnDeclaration: manually execute autocmds

gebner:
Without this change, upon calling EnDeclaration I lose both syntax highlighting and airline coloring in neovim 0.1.1.

AFAICT this happens because autocmds are [apparently not executed when using the API](https://github.com/neovim/neovim/issues/3853), so we have to call them manually.

I have tested this PR with both vim and neovim, and get syntax highlighting after EnDeclaration in both versions.  This might also solve @yazgoo's problems from #124.


vito-c:
LGTM!

yazgoo:
Thanks gebner!

Issue #151, opened 2016-02-03
Error with highlighting: EnErrorStyle undefined variable


jvican:
```Error detected while processing function ensime#au_cursor_moved..<SNR>31_call_plugin:
line   12:
E121: Undefined variable: g:EnErrorStyle
Press ENTER or type command to continue
Error detected while processing function ensime#au_cursor_moved..<SNR>31_call_plugin:
line   12:
E116: Invalid arguments for function matchadd(g:EnErrorStyle, '\%3l\%>22c\%<45c')
Press ENTER or type command to continue
Error detected while processing function ensime#au_cursor_moved..<SNR>31_call_plugin:
line   12:
E15: Invalid expression: matchadd(g:EnErrorStyle, '\%3l\%>22c\%<45c')
Press ENTER or type command to continue
Error detected while processing function ensime#au_cursor_moved..<SNR>31_call_plugin:
line   14:
E605: Exception not caught: vim.error:invalid expression```

Issue #152, opened 2016-02-03
Neovim import error when vim is executed and neovim module is not installed


jvican:
```Traceback (most recent call last):                                                                                                                            
  File "<string>", line 1, in <module>                                                                                                                        
  File "/Users/v751429/code/vim-config/bundle/ensime-vim/autoload/ensime.vim.py", line 19, in <module>                                                        
    from ensime_shared.ensime import Ensime                                                                                                                   
  File "/Users/v751429/code/vim-config/bundle/ensime-vim/ensime_shared/ensime.py", line 14, in <module>                                                       
    import neovim                                                                                                                                             
ImportError: No module named neovim                                                                                                                           
Error detected while processing function ensime#au_buf_enter..<SNR>66_call_plugin:                                                                            
line   14:                                                                                                                                                    
E605: Exception not caught: <type 'exceptions.NameError'>:name 'ensime_plugin' is not defined```

**Comment by alexDeCastroAtGit** [comment]
I'm getting the same error...

```

> Error detected while processing function ensime#com_en_add_import[1]..<SNR>28_ca
> ll_plugin:
> line   15:
> E605: Exception not caught: <type 'exceptions.NameError'>:name 'ensime_plugin' i
> s not defined

Do you remember what was done to recover from this error? I'm following the instructions on 
[(http://ensime.github.io/editors/vim/install/)]
but to no success...

```

ktonga:
Hey @alexDeCastroAtGit `ensime-vim` is a neovim remote plugin, so you need to run `:UpdateRemotePlugins`. Maybe that's your problem

PR #153, opened 2016-02-03
Fix issues #151 and #152

jvican:
These issues have been reported by @ceedubs and @milescrawford.
Thanks for the feedback, they were stupid errors that weren't
properly fixed from the last refactoring.


milescrawford:
This doesn't look like it's fixed for me.

I'm on c315b0755a6da4c92c265929830dbb0cce838364 in my bundle, and yet when I start vim I get:

Error detected while processing function ensime#au_cursor_hold..<SNR>32_call_plugin:
line   12:
E28: No such highlight group name: g:EnErrorStyle

PR #156, opened 2016-02-10
Fix EnClasspath problem and catch ws exceptions

jvican:
* Users couldn't execute it properly as a command

* Uses 'terminal' in nvim and '!' in vim to execute sbt, otherwise
nvim shows strange characters because the interactive session of sbt
is still opened and cannot be closed

Also, catch exceptions issued by the websockets module.


jvican:
review by @yazgoo

PR #158, opened 2016-02-13
Pep8

thisfred:
I added a few commands to the Makefile to help with (de)linting, and I did an initial run of `make autopep8` to clean up the style/formatting issues that were automatically fixable. After that I fixed some more of the reported lint issues by hand.

PR #159, opened 2016-02-13
made keyword arguments explicit

thisfred:
This passes through the keyword arguments explicitly so that typo's don't lead to hard to debug behavior but rather to syntax errors.

In one instance a `create` keyword argument was passed, which I assume was meant to be either `create_client` or `create_classpath`, but i'm not sure which, so I picked one, and added a comment. Please let me know either way, and I'll remove the comment and/or fix the usage.

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -824,20 +824,21 @@ class EnsimeClient(object):
             return result
 
 
-def execute_with_client(**kwargs):
+def execute_with_client(quiet=False,
+                        create_classpath=False,
+                        create_client=True):
     """Decorator that gets a client and performs an operation on it."""
     def wrapper(f):
-        # Gets the self reference and pass it
-        def wrapper2(self, *vimargs, **vimkwargs):
-            # By default, set to true if not present
-            if "create_client" not in kwargs:
-                kwargs["create_client"] = True
-            # All the missing parameters in
-            # kwargs will be False by default
-            client = self.current_client(**kwargs)
+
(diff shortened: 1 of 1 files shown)


ensime_shared/ensime.py
```diff
@@ -857,7 +861,9 @@ def client_keys(self):
 
     def client_status(self, config_path):
         """Get the client status of a given project."""
-        c = self.client_for(config_path, create=True)
+        # XXX: `create` wasn't used, should this be `create_client` or
+        # `create_classpath`?
+        c = self.client_for(config_path, create_client=True)
```
thisfred: this is the change I'm not sure about

jvican: This is a minor change that I did in my last PR. Could you please remove this change?

jvican:
I like the idea of passing explicit parameters but if the number of them starts to increase we may be better of going back to this solution. Please, remove that change I pointed out and rebase on top of the last PR. Then it'll get merged :+1:

ensime_shared/ensime.py
thisfred: so remove the create parameter altogether?

jvican:
Are you ready for merge @thisfred ?

thisfred:
yep, this one's ready too, but needs #158 to go in first

jvican:
Oops, actually you'll need to rebase this on top of the other PR so that it can get merged.

thisfred:
I redid the branch with just the one commit cherry picked.

jvican:
Thanks @thisfred!

PR #161, opened 2016-02-14
Add missing 'import json'

chessman:


jvican:
Thanks @chessman, That one slipped through the cracks!

Issue #169, opened 2016-02-27
complete_func does not work


ChrisCoffey:
I'm running `NVIM 0.1.2` with `+acl   +iconv    +jemalloc` with python 2.7.

Omni-completion appears to be broken on NeoVim in the current version. `fun_en_complete_func` in `ensime_shared.py` is dropping into the vim use-case, which in turn fails to call `complete()`. I played around with flipping the `if not base` condition, but it didn't seem to make a difference. 

Here's the log:
```
[...]

PR #172, opened 2016-03-04
This resolves several bugs related to omni-complete

ChrisCoffey:
This addresses #169 by correctly returning the results back to Vim as well as introducing a blocking `unqueue` flag for when we make synchronous calls to the ensime server.


jvican:
Thanks!

Issue #174, opened 2016-03-07
Support Public Symbol Searching


ChrisCoffey:
Often times I find that I want to look up a particular function or type by name (or at least the part of the name I remember) but I usually fall back on grep or ctags. I've been playing around with ENSIME's public symbol search and it seems to fit what I'm looking for. 

The only question I have left before pushing my changes for you guys to review is whether the results should go into the quickfix window or a side buffer. Do either @jvican or @yazgoo have a preference?

ches:
I'd like this feature too, the Atom ensime plugin implements it IIRC, with
a pop-up palette. I'd vote for quickfix window—if it's possible (now or in
the future) to scope the search to a single file then location list could
be used in that case and it will be nice and uniform.

It would also be nice if ensime-vim exposed this as an API that could be
hooked/called by Unite, CtrlP, etc.

>

jvican:
I see this feature in a quickfix window :smile:

ChrisCoffey:
@jvican ready for you to take a look. 

I did not expose it as an API, although I'm more than happy to do that in the future.

jvican:
Let's leave the public API for the future because it's something that it affects quite a lot of features of `ensime-vim`.

PR #175, opened 2016-03-08
Introduces EnSearch and machinery for adding quickfix items

ChrisCoffey:
This performs an ENSIME  `PublicSymbolSearchReq` and places the top 25 results into the quickfix window, creating a new list. I figured 25 was a pretty reasonable number, although the ENSME search result ordering leaves something to be desired, so its possible you won't have the symbol you're looking for.

The quickfix machinery itself should be useful for other navigation/ find usage commands once we're ready to add those as well.

To run this from NeoVim, make sure to run `:UpdateRemotePlugins`, otherwise `:EnSearch` will not appear as a command.

Addresses #174

### DIFF
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -82,6 +82,10 @@ function! ensime#com_en_type(args, range) abort
     return s:call_plugin('com_en_type', [a:args, a:range])
 endfunction
 
+function! ensime#com_en_sym_search(args, range) abort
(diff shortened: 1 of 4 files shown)


jvican:
Thanks ;)

ches:
Could we add a note to `doc/ensime-vim.txt` about this feature?

ches:
And thanks @ChrisCoffey for all the work you're contributing!

PR #176, opened 2016-03-08
Update to ensime website

ChrisCoffey:
Just completes the documentation migration over to ensime's website.


jvican:
Thanks a lot for this @ChrisCoffey, really appreciated! Could you please link directly to the `ensime-vim` doc? https://ensime.github.io/editors/vim/

PR #177, opened 2016-03-08
Fix start of autocompletion

jvican:
There was a bug that caused the omnicompletion to start the completion one column ahead. This PR fixes this and also adds no constraints to run autocompletion when the user asks for it. For instance, now we allow it even if you don't have written any seed.

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -841,11 +841,11 @@ class EnsimeClient(object):
         self.log("complete_func: in {} {}".format(findstart, base))
         if str(findstart) == "1":
             row, col, startcol = detect_row_column_start()
-            if startcol:
-                self.vim_command("write_file")
-                # Make request to get response ASAP
-                self.complete(row, col)
-                self.completion_started = True
+
+            self.vim_command("write_file")
+            # Make request to get response ASAP
+            self.complete(row, col)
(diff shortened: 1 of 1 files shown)


jvican:
Review by @ChrisCoffey.

jvican:
Please, LGTM as soon as you can as I want to merge this soon @ChrisCoffey :smile:

ChrisCoffey:
taking a look now

ensime_shared/ensime.py
```diff
@@ -833,23 +833,22 @@ def detect_row_column_start():
...
             start = col
             line = self.vim.current.line
-            # No completion from empty string
-            if start != 0 and line[start - 1] == " ":
-                start = -1
-            else:
-                while start > 0 and line[start - 1] != " ":
-                    start -= 1
-            return row, col, start + 1  # pos in no empty str
+            while start > 0 and line[start - 1] != " ":
```
ChrisCoffey: I forgot another common terminal case, the `.` operator. I think it probably makes sense for this condition to be:

```python
while start > 0 and (line[start -1] != " " and line[start - 1] != "."):
```

So we don't overwrite symbol names when chaining methods.

ChrisCoffey:
Other than my one comment this looks great

PR #178, opened 2016-03-08
Fix weird OSX folder rename error

jvican:
When we changed the configuration folder names of ensime-vim, we didn't
check if it was working for OSX. Apparently, the rename operation is
implemented in such a way in this OS that we cannot rename a non-empty
folder with `os.rename`, and we have to use `shutil.move` instead.


milescrawford:
Works for me! Ship.

Issue #180, opened 2016-03-09
:EnDeclarationSplit Does not Split The Buffer


ChrisCoffey:
@milescrawford mentioned that `:EnDeclarationSplit` was not splitting the buffer. We're overwriting the call options whenever we try to send a symbol at point req.

jvican:
Tracked by #181.

PR #181, opened 2016-03-09
Prevent overwriting call options

ChrisCoffey:
Pretty straightforward, looks like the refactor accidentally resulted in overwriting any previously set call options. This means that `split` is always missing, which breaks `:EnDeclarationSplit`.

addresses #180


jvican:
Good one! :smile: 
  
Could you please rewrite the commit with capital letter? It's the standard format for our commit history.

ChrisCoffey:
@jvican all set?

jvican:
Perfect!

PR #183, opened 2016-03-09
Perform dictionary get instead of index lookup. 

ChrisCoffey:
@jvican This fixes a bug I just introduces in `EnDeclaration`. Sorry about that.


jvican:
No problem! Thanks for pointing that out, I didn't notice it.

PR #185, opened 2016-03-10
Tighten up symbol search edge cases

ChrisCoffey:
Today I noticed an error when trying to search for a symbol from readonly files, which was happening because we were attempting to write the current buffer before searching. This was a misguided attempt on my part to ensure we don't jump away from edits.

I also found there are situations where ENSIME does not provide a position in the search results, which threw a key not found error. I added a guard against this as well.


jvican:
LGTM!

PR #186, opened 2016-03-10
Add Interface names to type inspection status line

ChrisCoffey:
This is a small change that causes `:EnInspectType` to display all of the types the inspected type extends/ implements.


jvican:
Nice, LGTM!

PR #189, opened 2016-03-11
Fix import suggestions for inner classes (with dollar sign)

kstep:
This change is [Review on Reviewable](https://reviewable.io/reviews/ensime/ensime-vim/189)


jvican:
LGTM.

Issue #192, opened 2016-03-14
Interested in seeing all the possible method signatures at typing time


casertap:
Hi guys,
I enjoy ensime-vim, very good work, thank you so much.

Is it possible to see all the possible method signatures at typing time?
For instance in the screenshot bellow I have the declaration on top and actorOf method has 2 signature:
```
actorOf(props: Props): ActorRef
actorOf(props: Props, name: String): ActorRef
```
in the buffer below I am creating a new actor using: system.actorOf...

![declaration](https://cloud.githubusercontent.com/assets/284394/13762333/5e33aaaa-ea92-11e5-8a47-99d4e40906f4.jpg)

At this point, I would like to see all the possible signatues.
Is this possible?

ChrisCoffey:
You can use the omni complete list, <C-x><C-o>, which pulls in what you
want.

casertap:
this is what I get by doing that:
![error](https://cloud.githubusercontent.com/assets/284394/13762668/9ae037dc-ea94-11e5-8bcf-b343670a7dfe.jpg)

ChrisCoffey:
Yikes. Can you link to the repo causing the issue? Also it would be helpful
if you can attach you server and ensime-vim logs. I'll take a look tonight.

casertap:
This is what is generated inside the ensime-vim.log
```
1457 2016-03-14 23:25:54.872751: 1.
1458 2016-03-14 23:25:54.873687: complete_func: in 1.
1459 2016-03-14 23:25:54.883299: complete: in
1460 2016-03-14 23:25:54.884252: 74 23
1461 2016-03-14 23:25:54.885065: 2021
1462 2016-03-14 23:25:54.885603: path: in
1463 2016-03-14 23:25:54.886156: send_request: in
1464 2016-03-14 23:25:54.886730: send: in
1465 2016-03-14 23:25:54.887448: send: {"callId": 18, "req": {"typehint": "CompletionsReq", "point": 2021, "caseSens": true, "maxResults": 100, "reload": false, "fileInfo": {"contents": "package com.packt.akka\n\nimport com.packt.akka.User\nimport scala.language.postfixOps\nimport akka.pattern.>
1466 2016-03-14 23:25:54.918334: 0 actorO
```
in server.log

```
[...]

PR #193, opened 2016-03-16
Vim passes a base parameter on the 0 callback

ChrisCoffey:
Turns out the Vim api switches between no `base` parameter for the initial call to `omniComplete`, and a present `base` for the callback. I was not handling the callback case correctly, and was dropping into the neovim handler, which resulted in the `IndexError`.

Addresses #192


casertap:
:+1: I need this

jvican:
Cool, thanks!

casertap:
This is beautiful man:
ensime

It is working. Good job.

PR #202, opened 2016-03-19
Remove cyclic method dependency and exit vim ASAP

jvican:


jvican:
Review by @ChrisCoffey

ensime_shared/ensime.py
```diff
@@ -136,16 +136,23 @@ def queue_poll(self, sleep_t=0.5):
 
         Value of sleep is low to improve responsiveness.
         """
+        connection_alive = True
         while self.running:
             if self.ws:
                 def logger_and_close(m):
                     self.log("Websocket exception: {}".format(m))
-                    self.teardown()
+                    if not self.running:
```
ChrisCoffey: Can this condition ever be hit? on the first exception, `self.running= True`, which means we call `teardown()`, which flips `self.running= False`. This is all during the same pass. On the next pass, the `self.running` condition will evaluate to `False`, terminating the loop, which prevents any future evaluations of `logger_and_close()`.

ensime_shared/ensime.py
```diff
@@ -136,16 +136,23 @@ def queue_poll(self, sleep_t=0.5):
 
         Value of sleep is low to improve responsiveness.
         """
+        connection_alive = True
```
ChrisCoffey: I'm not totally sure why this is necessary. From my reading of the code, `teardown()` will flip `self.running` to `False`, which should terminate the loop on the next pass. So this saves a 1/2 second sleep, but that's about it. 

Am I correct in this interpretation or is there something I'm missing?

ChrisCoffey:
@jvican if you're going to use this approach with a local instead of global variable, it may make sense to completely remove `self.running` and only use `connection_alive` in `queue_poll`.

ensime_shared/ensime.py
jvican: Yes, it can be hit. It's hit every time there's a Websocket exception and we haven't called `teardown`, that is, when an unexpected websocket exception happens and the user hasn't exited from vim. In that case, `self.running` is still `True` and we need to close the editor (we choose to). Otherwise we mark the connection as dead and exit without executing `time.sleep`.

jvican:
We can't do that as we still need a way to stop the thread from the actual vim proxy through a global variable. Also, they represent different things. `self.running` represents that `ensime-vim` is running while `connection_alive` represents that the websocket thread is. I'd keep them apart to make that clear.

ensime_shared/ensime.py
ChrisCoffey: I'm still pretty curious about this. I ran a quick test where I load up a file, let the server boot, kill the server, then issue an command. Here are the logs for current mater & your branch:

Master
```
2016-03-20 16:44:25.443928: Websocket exception: Connection is already closed.
2016-03-20 16:44:25.444037: teardown: in
2016-03-20 16:44:30.522621: type: in
2016-03-20 16:44:30.523756: send_at_position: in
2016-03-20 16:44:30.524753: path: in
2016-03-20 16:44:30.525031: 56 11
2016-03-20 16:44:30.525443: 1944
2016-03-20 16:44:30.525521: send_request: in
2016-03-20 16:44:30.525715: send: in
2016-03-20 16:44:30.525868: send: {"callId": 2, "req": {"range": {"to": 1947, "from": 1944}, "typehint": "TypeAtPointReq", "file"
: "/Users/ccoffey/workspace/open_source/ensime-server/.ensime_cache/dep-src/source-jars/org/apache/lucene/document/TextField.java
"}}
2016-03-20 16:44:30.526153: send error: socket is already closed., reconnecting...
```

remove-cycle-teardown
```
2016-03-20 17:01:36.672266: Websocket exception: Connection is already closed.
2016-03-20 17:01:36.672395: teardown: in
2016-03-20 17:01:57.979673: open_declaration: in
2016-03-20 17:01:57.980702: 74 45
2016-03-20 17:01:57.981463: 3175
2016-03-20 17:01:57.981560: path: in
2016-03-20 17:01:57.982108: send_request: in
2016-03-20 17:01:57.982248: send: in
2016-03-20 17:01:57.982375: send: {"callId": 3, "req": {"typehint": "SymbolAtPointReq", "file": "/Users/ccoffey/workspace/open_source/ensime-server/core/src/main/scala/org/ensime/indexer/IndexService.scala", "point": 3176}}
2016-03-20 17:01:57.982701: send error: socket is already closed., reconnecting...
```

In both cases it looks like the teardown code fires and jumps out of the loop on the next cycle since we only see the `Websocket exception` message once in each branch, so I went a little further and added a log line on line `144`, right above `connection_alive = False`. I never saw that log line get hit. 

Is killing the server too extreme of a test in this case, or is there something else I'm missing?

ChrisCoffey:
@jvican how do we use `self.running` to stop the vim proxy? I don't see it referenced anywhere in the plugin code if I run `grep -R 'running' ./` from the project root; is there some implicit usage in the Vim api?

jvican:
We don't stop the vim proxy. We stop the websocket thread from the vim proxy (or vim plugin) through `self.running = False`. If we have only one local variable as you propose, you can't anymore stop that thread. And, also, it's less explicit and precise. I think it's a good idea to make the concepts of the code clear since they represent different things.

ensime_shared/ensime.py
jvican: Killing the server is too extrem here because I think that we're checking if the server is alive and when it's not we explicitly call `teardown` as you see in the logs `teardown: in`. The statement we're discussing is covering the case in which teardown is not called and that may happen because of other spurious errors. My take is that the ws thread and the plugin are independent and, therefore, the ws thread may catch an exception when the vim plugin hasn't exited. Does that make sense?

ChrisCoffey: Yup, that makes sense to me. 

The code doesn't reflect that approach though, because we still fire `logger_and_close()` on any exception, which in turn is guaranteed to call `teardown()` the first time (because we're still `running`), which is guaranteed to set `self.running = False`, which in turn ends the loop and complete's the thread. 

For the approach you're talking about- which I'm all for - I think we need to be more selective about which exceptions we catch with `teardown()`.

jvican: Show me some suggestions and I'll be pleased to add them to the PR. Specifically, what do you mean with being more selective?‎

I'm thinking that maybe a good idea would be not to close and exit vim if such a `WebsocketException` happens and, instead, we just disable `ensime-vim` (via runtimepath) and notify the user. This is probably the cleanest solution.

ChrisCoffey: I think disabling the plugin on any `WebsocketException` is a great solution!

jvican: You'll have a PR tonight and hope we merge it today! :+1:

jvican:
I'll get back to this hopefully today, I've had some rather busy days lately!

jvican:
Have a look at my changes @ChrisCoffey

ensime_shared/ensime.py
```diff
@@ -187,40 +203,89 @@ def ready_to_connect():
...
+        """Warn users that a module is not available in their machines."""
         msg = feedback["module_missing"]
         self.raw_message(msg.format(name, name))
 
+    def get_runtime_path(self):
+        """Return folder in which the `ensime-vim` plugin is locate."""
+        # Embed path, is there a smarter way to do it?
+        # Note that the path has to be exactly the same
+        # than what 'set runtimepath?' returns
+        return "~/.nvim/plugged/ensime-vim/"
```
jvican: This will obviously change for people running vim instead of neovim. In my current install, vim still points to the same dir, but if one doesn't have it installed it should change...

ChrisCoffey: Ideally this should pull from a variable we can set from our `.vimrc`. For instance, I run out of `~/.config/nvim/plugged/ensime-vim` rather than `~/.nvim/`

ChrisCoffey: or use vim itself to resolve where the current script is executing

ChrisCoffey:
This looks fine except for the hardcoded path. Once that's fixed it should be good to merge.

ensime_shared/ensime.py
jvican: It's not that easy since there's no way to get the runtimepath from vim, only to show it to the user via `set runtimepath?`. The main problem is that the path has to be exactly the same that it resides in runtimepath so that it's effectively removed from it.

ChrisCoffey:
lgtm :+1:

PR #204, opened 2016-03-22
Copy over server issue template

ChrisCoffey:
I like the level of specificity the server template pushes folks to provide, and I'd love to see the same thing here.


jvican:
LGTM

Issue #190, opened 2016-03-13
Package Inspector


ChrisCoffey:
It looks like ENSIME & the Emacs plugin have support for displaying an outline of a package's contents. Is this something we want to bother porting to ENSIME-Vim, or are we happy enough with ctags and other outlining options? I could go either way on this.

jvican:
I think it's good if we don't rely on ctags for this and we do port it to ensime-vim :+1:

ChrisCoffey:
Works for me. I'm thinking the output should write to a narrow vertical buffer, similar to nerd tree. Does that sound reasonable to you @jvican?

jvican:
Perfectly reasonable @ChrisCoffey!

PR #206, opened 2016-03-23
Introduces basic Package Inspector

ChrisCoffey:
This introduces a very simple package inspector. Users provide a full package name, and the resulting class names are added to a new vertical split.

There are a few obvious enhancements for things like jumping to the declaration, finding all usages (via grep), displaying symbol type, coloring results, etc... that I think would be really useful either as community additions or things I'll get to when I have the opportunity.

#190


jvican:
Good job, like it! LGTM

Issue #211, opened 2016-03-28
Upgrade debugging implementation


ChrisCoffey:
@2chilled reported that our debugging implementation is outdated to the point of not being supported by the current version of ensime-server. Turns out we're sending `DebugStartReq` to the server instead of `DebugAttachReq`, which prevents the debugger from correctly launching.

PR #214, opened 2016-03-29
Support new remote debugger

ChrisCoffey:
This flips the debug api over to use `DebugAttach` & updates the doc. 

Addresses #211


ensime_shared/ensime.py
```diff
@@ -744,15 +748,16 @@ def clear_breaks(self, args, range=None):
 
     def debug_start(self, args, range=None):
         self.log("debug_start: in")
-        if len(args) > 0:
+        if len(args) > 1:
```
2chilled: Can we make the args optional? In most cases you want to debug localhost on port 5005 I think

jvican: Yup, definitely up for a default port number :+1:

jvican:
LGTM :+1:, but the default port number is worth a shot! Thanks!

ChrisCoffey:
sounds great to me!

Issue #207, opened 2016-03-23
Include Symbol Type in Package Inspector


ChrisCoffey:
`PackageInfo` objects from Ensime contain information on whether a symbol is a class, trait, etc... We should include this information in the package inspector's output.

PR #215, opened 2016-03-30
Add type information to package members

ChrisCoffey:
This addresses #207. New types should be added to syntax support as they appear.

This also flips the inspector to a scratch buffer rather than standard buffer.


jvican:
LGTM :+1:

Issue #208, opened 2016-03-23
Open Symbol Declaration From Package Inspector


ChrisCoffey:
The Package Inspector contains all of the top-level symbols defined within a package, so we should allow users to jump directly to the symbol declaration from the inspector. I'm thinking we can do this by performing a `SymbolByNameReq` then opening the resulting file in a new buffer.

Thoughts?

jvican:
Yes, that's how I'd do it.

PR #217, opened 2016-04-02
Introduces SymbolByName and a Jump-To-Package Symbol

ChrisCoffey:
This introduces two new commands, `:EnSymbolByName` and `:EnPackageDecl`. The documentation explains how to access these commands. 

`:EnPackageDecl`'s implementation relies on knowing the structure of the package tree as follows:

```
org.ensime
  Class: BuildInfo
  Object: BuildInfo$
  : api
    Class: AddImportRefactorDesc
    Object: AddImportRefactorDesc$ < Cursor is here
  ...
```

With the cursor on `Aird$`, I pull the lines back to the start then walk the tree up to the root & reconstruct the FQN (`org.ensime.api.AddImportRefactorDesc$`). I then send the FQN to the server via a `SymbolByNameReq` & open a new vertical window with the declaration inside. The new buffer only opens when we move back to a `.scala` file, so that's worth looking into in the near future.

Addresses #208


ChrisCoffey:
I don't love that this uses commands & I'm open to an alternative way of exposing the feature.

jvican:
This is nice, I like the idea! LGTM

Issue #168, opened 2016-02-25
Execute :EnClasspath to set a classpath (on master)


casertap:
When I open a .scala file in vim, I get the bip sound at regular interval and I get the message:
Execute :EnClasspath to set a classpath
my .ensime looks like this (ignore the line numbers, lines are truncated (>) at the end because I could not do a proper copy paste from the vim of my docker machine):
```
1 (
  2  :root-dir "/code/scala-sbt-template"
  3  :cache-dir "/code/scala-sbt-template/.ensime_cache"
  4  :scala-compiler-jars ("/root/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.10.3.jar" "/root/.ivy2/cache/org.scala-lang/scala-library/jars/sca>
  5  :name "scala-sbt-template"
  6  :java-home "/usr/lib/jvm/java-7-openjdk-amd64"
  7  :java-flags ("-Xss2m" "-Xms1024m" "-Xmx1024m" "-XX:ReservedCodeCacheSize=128m" "-XX:MaxPermSize=256m")
  8  :reference-source-roots nil
  9  :scala-version "2.10.3"
 10  :compiler-args ("-unchecked" "-deprecation" "-feature" "-Xlint" "-Yinline-warnings" "-Yno-adapted-args" "-Ywarn-dead-code" "-Ywarn-numeric-widen" "-Xfuture")
 11  :formatting-prefs nil
 12  :disable-source-monitoring nil
 13  :disable-class-monitoring nil
 14  :subprojects ((
 15    :name "scala-sbt-template"
 16    :source-roots ("/code/scala-sbt-template/src/main/scala" "/code/scala-sbt-template/src/test/scala")
 17    :targets ("/code/scala-sbt-template/target/scala-2.10/classes")
 18    :test-targets ("/code/scala-sbt-template/target/scala-2.10/test-classes")
 19    :depends-on-modules nil
 20    :compile-deps ("/root/.ivy2/cache/org.scala-lang/scala-library/jars/scala-library-2.10.3.jar")
 21    :runtime-deps nil
 22    :test-deps ("/root/.ivy2/cache/org.ow2.asm/asm/jars/asm-4.1.jar" "/root/.ivy2/cache/org.ow2.asm/asm-analysis/jars/asm-analysis-4.1.jar" "/root/.ivy2/cache/or>
[...]

PR #218, opened 2016-04-06
Double Quote FeedKeys expr-quote \e for Error Bell

ChrisCoffey:
Looks like the quotes were flipped for the escape character. I haven't been able to test this in Vim yet (my docker env is borked).


jvican:
Thanks for your work @ChrisCoffey, LGTM!

ches:
Boom, I just tested cherry-picking this commit before @jvican merged it, and it fixes the issue on #168 for me. Thanks!

PR #219, opened 2016-04-06
Fix missing :EnPackageDecl declaration for Vim

ches:
Pressing `<Space>` in the package inspector window was resulting in an error in regular Vim because `: EnPackageDecl` was undefined. This looks to have been a copy-paste oversight in #217, including two copies of `:EnSymbolByName` by accident but having it correct in the Neovim decorators.

Also adds a note to the help to try to reduce the potential confusion of the split not being immediately opened in current implementation. (Funnily, I missed this caveat in the original PR, and it turns out that the splits are queued up… so if you try `<Space>` four times and then move the focus, you get four splits 😇   Definitely room for improvement but I like the start).

cc @ChrisCoffey


ChrisCoffey:
Looks good, thanks for the PR & sorry for the typo!

@jvican merge when you're ready.

jvican:
LGTM!

PR #221, opened 2016-04-06
Do not write file when invoking completion

### LINKED ISSUE #220: Don't write file when invoking completion
Currently `complete_func` invokes `self.vim_command("write_file")` when starting completion—is this necessary?

My assumption is that it was done either to try to get the most up-to-date accurate state of source possible, and/or to be extra cautious in case of a crash being caused. But it's super annoying if, say, you're using sbt triggered execution: compiles are certainly going to fail when I'm in the middle of writing a statement 😄   Considering how frequently this will happen (especially with YouCompleteMe automatically invoking `omnifunc` on-the-fly e.g. when I type `.`), it's really noisy and probably quite a drain on laptop batteries…

I could submit a PR removing the `"write_file"` but wanted to make sure I understood the context.

ches:
Closes #220

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -1082,7 +1082,6 @@ class EnsimeClient(object):
         if str(findstart) == "1":
             row, col, startcol = detect_row_column_start()
 
-            self.vim_command("write_file")
             # Make request to get response ASAP
             self.complete(row, col)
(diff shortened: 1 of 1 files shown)


ChrisCoffey:
:+1: 

@jvican already weighed in, merging.

PR #226, opened 2016-04-06
Try to use a browser with no required configuration

ches:
It would be a nicer experience if there was nothing to configure for the `:EnDocBrowse` command, so it would work out of the box. [Python's `webbrowser` module][1] generally does a good job of opening the user's default browser across platforms. It supports the `BROWSER` environment variable so that still allows for customization and should be mostly backward-compatible with the current approach.

I would also like to propose moving the `self.raw_message(prefix_msg.format(url))` to only run in the unsuccessful case (which should hopefully be very unlikely after this change)—it causes the user to have to press an acknowledgment key to return to editing after returning from their browser window, and IMO this is a needless interruption.

[1]: https://docs.python.org/2/library/webbrowser.html

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -1,6 +1,7 @@
 import sys
 import os
 import inspect
+import webbrowser
 
 # Ensime shared imports
 from ensime_shared.error import Error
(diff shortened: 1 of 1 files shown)


ches:
I went ahead and moved the `raw_message` as I mentioned. If there is disagreement about that I can remove that commit.

ChrisCoffey:
:+1:

jvican:
LGTM :+1: Please, could you please splash the commits?

ches:
Squashed.

jvican:
I'm merging this optimistically but I hope that it can handle scripts that invoke the browser as well!

ches:
For future reference in case anyone wishes to do it, I tried this successfully on OS X:

    $ export BROWSER="/Users/ches/bin/vim-browser %s"
    $ chmod a+x ~/bin/vim-browser

Contents of `vim-browser`:

```sh
#!/bin/sh

exec open -a Safari "$@"
```

`:EnDocBrowse` then opened in Safari (my default browser is Chrome). AFAICT the script path(s) need to be absolute in the `BROWSER` value, i.e. no expansion of `~`.

jvican:
Thanks for the reference and the snippet @ches.

Issue #225, opened 2016-04-06
EnSymbol does not print results


ChrisCoffey:
### Steps to reproduce

1. Open a scala file
2. Run EnSymbol

### Expected behaviour
Should print the file path for the symbol under cursor

### Actual behaviour
nothing happens

PR #227, opened 2016-04-07
Implement display side of EnSymbol

ChrisCoffey:
Adds the new `display` option for symbol handling & sets this flag for `EnSymbol`. It's just a simple echo.

Addresses #225


jvican:
LGTM :smile:

Issue #223, opened 2016-04-06
Package Inspector Does not open symbol immediately


ChrisCoffey:
The Package Inspector does not actually open the targeted symbol until the user moves to another buffer. It should instead open a new buffer & jump to it as soon as the results are returned from the server.

ches:
See #217 and #219 for some context/history.

PR #228, opened 2016-04-07
Implements expected symbol opening from Packages

ChrisCoffey:
It took a bit more exploring than I expected to get the synchronous behavior that I wanted, but I'm pretty happy with the final product. Upon hitting `<Space>` a request for the symbol under cursor is sent to the server & the client will wait a few seconds for the response. 

The response file is opened in a new vertical buffer immediately to the right of the package inspector.

An alternative solution to calling `unqueue(shouldWait=true)` would have been to add the `CursorHold` handler to the `package_info` filetype, but that felt unclear as a slow response would let users hop back to their previous buffer and move around before being pulled to the new window.

Addresses #223

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -80,6 +80,7 @@ class EnsimeClient(object):
             self.vim_command("set_updatetime")
             self.vim_command("set_ensime_completion")
             self.vim.command("autocmd FileType package_info nnoremap <buffer> <Space> :EnPackageDecl<CR>")
+            self.vim.command("autocmd FileType package_info  setlocal splitright")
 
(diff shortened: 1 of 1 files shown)


ensime_shared/ensime.py
```diff
@@ -1095,7 +1097,7 @@ def detect_row_column_start():
             if self.completion_started:
                 self.vim_command("until_first_char_word")
                 # Unqueing messages until we get suggestions
-                self.unqueue("", timeout=self.completion_timeout, should_wait=True)
```
jvican: Thanks for removing this, I completely forgot.

jvican:
LGTM :+1:

PR #229, opened 2016-04-07
Show friendly error message for PackageInspector w/o fqn

ChrisCoffey:
Not much to this one, just adding a quick error message explaining how to use the package inspector.

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -749,6 +749,10 @@ class EnsimeClient(object):
             "file": self.path()})
 
     def inspect_package(self, args):
+        if not args:
+            msg = commands["display_message"].format("Must provide a fully qualifed package name")
+            self.vim.command(msg)
+            return
         self.send_request({
             "typehint": "InspectPackageByPathReq",
             "path": args[0]


ches:
This would be good for when any invalid argument is given regardless, but I was thinking it would be nice if the plugin tried to find and use the package of the current buffer's file if the argument is empty. That's often what you want and typing FQNs is not much fun 😄

ChrisCoffey:
Totally agree, typing FQNs is awful. I believe we do notify if the name is invalid, but it occurs async. For this, we could use a regex but I didn't see much of a point since you'll receive feedback from the server that the package doesn't exist.

I really like the idea of extracting the current file's package in the case no argument is provided. Shouldn't take too long to do.

I'm also toying with prompting rather than requiring an argument with the command. Thoughts?

ches:
Just my opinion, but I feel arguments make the commands more flexible and
fluid to use than prompting (scriptable/composable/etc.). Can't recall
which it is because I'm not at my computer now, but I had logged a TODO
about one of the existing prompts because it was inconsistent with others
using arguments. For instance if I want a prompt-like experience for
something I use often, I can choose to create a mapping that fills the
command line without a `<CR>` included.

Command args can take advantage of completion too, which would probably be
too slow with server calls but there may be cases where caches could fit in
the future.

ChrisCoffey:
That's a very good point, which I agree with :).

We should have a conversation about standardizing on either args or prompts.

jvican:
Sorry, I don't follow what @ches meant. What prompts are you talking about? Could you please elaborate on that?

ches:
@jvican The example I was thinking of is `:EnSearch`. My instinct is to do `:EnSearch SomeSymbolName` but currently you can't use it that way, you only enter `:EnSearch` and press <kbd>Enter</kbd>, and then ensime-vim gives you an input prompt to enter the symbol name. This is inconsistent with how `:EnShowPackage` and `:EnSymbol` currently work, possibly others.

`:EnSearch` is a good example IMO of how arguments might leave more flexibility for the user and feel like more idiomatic Vim to me: if it used an argument, I could create my own mapping to do `:exec 'EnSearch ' . expand('<cword>')<CR>` to search for word under cursor if that is something I want to do often.

ches:
(It might even be nice if it was a built-in feature that `:EnSearch` used `<cword>` automatically if no argument is given, since that feels similar to how `:EnSymbol` works, but I hope my point about user scriptability is still clear).

ChrisCoffey:
Let's continue this conversation on #232 rather than the pr comments. I'll move over the conversation thus far.

ches:
So rounding up: right now I think the change in this PR is a welcome improvement already, since currently you get an ugly Python exception if you use the command incorrectly. 👍  from me FWIW.

I created #233 to track the idea of making `:EnShowPackage` use the current file/scope's package by default, as a later enhancement whenever someone has time to implement. And the UX discussion about arguments vs. prompts that has wider scope is now on #232.

ChrisCoffey:
Merging this now.

Issue #194, opened 2016-03-17
ensime-gradle + ensime-vim incompatibility


codyrioux:
Using the latest gradle integration located on https://ensime.github.io I'm running into an issue in which I get an error about the property "scala-version" not existing. Python is throwing this specifically because it is missing from the config map loaded in the loader.py
```
Error detected while processing function ensime#au_buf_enter[1]..<SNR>116_call_plugin:
line   15:
E605: Exception not caught: <type 'exceptions.KeyError'>:'scala-version'  
```

If I generate the .ensime file using sbt (same dependencies) I get a much different sized file, and it functions. The second solution is to edit `ensime_shared/launcher.vim` line 223 from:

```python
            #original
            "(", " ").replace(")", " ").replace('"', "").split(" :")
            #new
            "(", " ").replace(")", " ").replace('"', "").split(":") #space before colon removed
```

**Versions**
Gradle: Gradle 2.10
SBT: 0.13
vim: VIM - Vi IMproved 7.3 (2010 Aug 15, compiled Oct 27 2015 16:22:14)
macvim: VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Mar 16 2016 15:52:28)

Regardless of wether I use SBT or have made the edit to the launcher.py I'm getting a message on my vim status line:
```
not found: \e
```

It appears to be refreshing, so I suspect the cause is running in a loop.
Upon quitting I'm hanging on the message and have to force quit vim:
```
:EnClasspath to set a classpath
```

I suspect the initial issue is with the ensime-gradle plugin, but I thought I'd check here about the rest first. If I can be of any help by writing documentation after this, or even a pull request after getting some input I'm certainly willing to do so.

codyrioux:
Gist with the generated .ensime files [here](https://gist.github.com/codyrioux/223b798cfdd1b61e01b4).

ChrisCoffey:
If you're up for extending the launcher to support both plugins that's great! From the docs it looks like the original code was built to support only sbt, hence the break you're seeing. 

As for the looping issue, could you take a look through your `.ensime-vim` log & post the 150 or so lines before you killed vim to a gist?

codyrioux:
Hey Chris,

[...]

PR #235, opened 2016-04-08
Use a proper S-expression parser for .ensime

ches:
This uses [a real S-expression parser][1] for `.ensime`, hopefully a step toward being more robust for the output of other build tools like Gradle (see #194).

This is just a sketch to get feedback on the idea—since `sexpdata` is a reasonably small module that's self-contained in a single file, and its license allows it, I just vendored the source code so we don't have to ask users to `pip install` another dependency. But of course I'm open to feedback on that.

It'd be nice to organize the nested functions somewhere or maybe make an `EnsimeConfig` dict wrapper class or something for them—that's why I've preemptively made separate functions with docstrings—but I can go further with the organization after getting your thoughts.

/cc @codyrioux

- [ ] Refactor the `disable_plugin` machinery from `EnsimeClient` to be usable in launcher
- [ ] Update documentation site to add `sexpdata` to the pip installs

[1]: https://github.com/tkf/sexpdata


ches:
By the way, my recursive dict-transforming is fairly quick and dirty—if the unit tests were working currently it'd be nice to test a few normal `.ensime` and funkier S-expression structures… I'll try to get around to getting unit tests going again too, but in the meantime if folks can try this with a few different project `.ensime`s and build tool plugins that'd be a good place to start as far as more testing goes.

ChrisCoffey:
Great idea, I'll give this a look tonight

ChrisCoffey:
I would rather use the pip requirements file than vendor the code. That way it's still just one thing for users to run. 

If it feels to difficult to ask people to run `pip install -r <path-to-plugin>/requirements.txt` then we could go so far as to bootstrap the install.

jvican:
Thanks for bringing this @ChrisCoffey, I completely agree on that. To be honest, I don't want hardcoded vendor software in this repository, I'd advocate for pip.
  
Otherwise, it's a great idea to use S-Parser, thanks for the contribution @ches.

ches:
Okay, I've quickly removed the vendored `sexpdata` copy so that we can move past that point of discussion. This really should probably use `module_exists` and something like the `disable_plugin` machinery that is in `EnsimeClient` for a nicer user experience if the module isn't installed properly, but refactoring that to be usable to the launcher is something I'll have to come back to when I have a few more minutes to spare.

Updating docs to include the new pip install will have to be a separate pull request to the site repo when it's ready to be accepted, I guess. I'll squash commits on this branch when it's ready.

These todos are now reflected in the PR description.

ches:
FWIW as long as it's only a couple or a few pip modules my suggestion is that we just advise `pip install websocket-client sexpdata` in the documentation, not introduce a requirements file that ordinary users need to be aware of—not all Vim users are familiar with details about Python packaging conventions, and we'll get into the mess of having to explain how the file path may differ between the umpteen plugin managers used across the Vim ecosystem 😩 

IMO the fewer particularities we need to explain that go beyond standard pure-VimL plugins, the better.

ChrisCoffey:
Agreed, the more I thought about using a reqs file, the less I liked it. 

If you can update the documentation on the ensime page I'm happy to land this. Could you just cross-reference the documentation pull here?

ChrisCoffey:
Made the doc change. Merging.

Issue #240, opened 2016-04-25
EnShowPackage Throws Exception


ChrisCoffey:
the `append_line` command shortcut was mistakenly removed in a recent pull. This breaks the show package functionality.

ChrisCoffey:
Introduced by #230 , probably a typo.

Issue #239, opened 2016-04-24
EnPackageDecl is not documented


yyadavalli:
The command `EnPackageDecl` is not documented. In my case its not working, so not able to figure out what it is doing.

ches:
This is basically the same issue as what #231 now represents, but that one ought to be renamed to make that clearer—I'll do that now.

Issue #231, opened 2016-04-08
Make implementation details of Package Inspector private, instead of using commands


ches:
The Package Inspector feature has used two commands which are essentially internal implementation details, `:EnPackageDecl` and `:EnSymbolByName` (see #217). These lead to confusion like #239, and the below struck-out stuff of my own.

I think we could support the feature with greater encapsulation. My initial thought was using script-scoped functions (`s:foo`), exposed/called as a `<Plug>` mapping in the case of `:EnPackageDecl`.

The functionality of `:EnSymbolByName` might make sense to leave as a "public" function (the current `ensime#com_en_symbol_by_name`) and maybe even leave the command too (but document it), but I'm not sure how it practically differs from `:EnDeclarationSplit` in use case, I haven't closely compared the code for the two yet. Maybe a refactor could combine them?

~~I just realized that this command exists but isn't documented. It doesn't seem to work.~~

### ~~Steps to reproduce~~

~~1. Run `:EnSymbolByName something`~~
~~2. Wait for a response~~

### ~~Expected behaviour~~

~~I think the intended behavior is to echo where the symbol is defined, like `:EnSymbol` does but taking an arbitrary argument instead of using `<cword>`.~~

### ~~Actual behaviour~~

~~Nothing happens.~~

ches:
Forgot to note in the issue body: I'm guessing this command might need to be wired up like #227 did.

ChrisCoffey:
It's actually not intended to work like `:EnSymbol` but rather like ':EnDeclarationSplit v', so this is more of an issue with the naming than anything else. 

I added this to support the Package Inspector's jump-to-definition

ches:
Oh right, I completely forgot (and wasn't looking at the source) that it was an implementation detail of the Package Inspector. Could we just use a function for that (possibly private, `s:foo`), instead of exposing a command? I know you had reservations about that before. This was the case that made me think about instead exposing a `<Plug>` mapping to such a function, so users could change the mapping without a command being needed.

When I can make time I'll try to work up a pull request to show what I mean.

ChrisCoffey:
That'd be great. I'll take a stab at this as well later tonight.

PR #241, opened 2016-04-25
Change EnPackageDecl to a function

ChrisCoffey:
This removes the `EnPackageDecl` wart from the plugin by moving it into a function where it belongs. I'm curious whether you ( @ches & @jvican ) think this solution is adequate. 

I also restored the `append_line` cmd to the plugin's command lookup, looks like it was removed as part of the import suggestions.

Addresses #240, #239 & #231.

### DIFF
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -99,8 +99,8 @@ function! ensime#com_en_declaration_split(args, range) abort
     return s:call_plugin('com_en_declaration_split', [a:args, a:range])
 endfunction
 
-function! ensime#com_en_package_decl(args, range) abort
-    return s:call_plugin('com_en_package_decl', [a:args, a:range])
+function! ensime#fun_en_package_decl(args, range) abort
+    return s:call_plugin('fun_en_package_decl', [a:args, a:range])
 endfunction
 
 function! ensime#com_en_symbol_by_name(args, range) abort
--- a/plugin/ensime.vim
+++ b/plugin/ensime.vim
@@ -19,7 +19,6 @@ command! -nargs=* -range EnShowPackage call ensime#com_en_package_inspect([<f-ar
 command! -nargs=* -range EnDeclaration call ensime#com_en_declaration([<f-args>], '')
 command! -nargs=* -range EnDeclarationSplit call ensime#com_en_declaration_split([<f-args>], '')
 command! -nargs=* -range EnSymbolByName call ensime#com_en_symbol_by_name([<f-args>], '')
-command! -nargs=* -range EnPackageDecl call ensime#com_en_package_decl([<f-args>], '')
 command! -nargs=* -range EnSymbol call ensime#com_en_symbol([<f-args>], '')
 command! -nargs=* -range EnRename call ensime#com_en_rename([<f-args>], '')
 command! -nargs=* -range EnInline call ensime#com_en_inline([<f-args>], '')
@@ -38,6 +37,10 @@ command! -nargs=* -range EnToggleFullType call ensime#com_en_toggle_fulltype([<f
 command! -nargs=* -range EnOrganizeImports call ensime#com_en_organize_imports([<f-args>], '')
 command! -nargs=* -range EnAddImport call ensime#com_en_add_import([<f-args>], '')
 
+function! EnPackageDecl() abort
+  return ensime#fun_en_package_decl()
+endfunction
+  
 function! EnCompleteFunc(a, b) abort
     return ensime#fun_en_complete_func(a:a, a:b)
 endfunction
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -63,6 +63,7 @@ commands = {
     "syntastic_show_notes": "silent SyntasticCheck ensime",
     "get_cursor_word": 'expand("<cword>")',
     "select_item_list": 'inputlist({})',
+    "append_line": 'call append({}, {!r})'
 }
 
 
@@ -78,7 +79,7 @@ class EnsimeClient(object):
             self.vim_command("highlight_enerror")
             self.vim_command("set_updatetime")
             self.vim_command("set_ensime_completion")
-            self.vim.command("autocmd FileType package_info nnoremap <buffer> <Space> :EnPackageDecl<CR>")
+            self.vim.command("autocmd FileType package_info nnoremap <buffer> <Space> :call EnPackageDecl()<CR>")
             self.vim.command("autocmd FileType package_info  setlocal splitright")
 
         def setup_logger_and_paths():
@@ -1269,7 +1270,7 @@ class Ensime(object):
         client.symbol_by_name(args, range)
 
     @execute_with_client()
-    def com_en_package_decl(self, client, args, range=None):
+    def fun_en_package_decl(self, client, args, range=None):
         client.open_decl_for_inspector_symbol()
 
     @execute_with_client()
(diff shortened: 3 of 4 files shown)


plugin/ensime.vim
```diff
@@ -38,6 +37,10 @@ command! -nargs=* -range EnToggleFullType call ensime#com_en_toggle_fulltype([<f
 command! -nargs=* -range EnOrganizeImports call ensime#com_en_organize_imports([<f-args>], '')
 command! -nargs=* -range EnAddImport call ensime#com_en_add_import([<f-args>], '')
 
+function! EnPackageDecl() abort
+  return ensime#com_en_package_decl()
+endfunction
```
ches: Typo here: you forgot to rename the `com_` to `fun_`.

I know you're following patterns that are already there, but this leads me to ask a question or two I've wondered about for awhile: do we need the boilerplate of this extra wrapper function, instead of just invoking `ensime#fun_en_package_decl()` directly? If we end up having to put tons of wrappers in `plugin/` then we won't get much benefit out of autoload.

Perhaps related, it always confuses me that we have many functions (and commands) taking `(args, range)` that don't actually use them. I assume without reading closely right this moment that this is baked into expectations of the `call_plugin` delegation, and possibly the `@execute_with_client()` decorator too. It seems like a smell to me, hard to understand and prone to maintenance headaches. It'd be a big refactoring to change this, and obviously out of scope for this pull request… Just wanted to bring it up since I keep wondering. Perhaps we can discuss/track it in a separate discussion or sync up on Gitter at some point.

In other news, it has finally occurred to me what the `com_` prefix means 😊

autoload/ensime.vim
```diff
@@ -91,8 +91,8 @@ function! ensime#com_en_format_source(args, range) abort
     return s:call_plugin('com_en_format_source', [a:args, a:range])
 endfunction
 
-function! ensime#com_en_declaration(args, range) abort
-    return s:call_plugin('com_en_declaration', [a:args, a:range])
+function! ensime#fun_en_declaration(args, range) abort
+    return s:call_plugin('fun_en_declaration', [a:args, a:range])
```
ches: Actually isn't `ensime#com_en_package_decl` what you meant to rename here, not `com_en_declaration`?

ches: Actually isn't `ensime#com_en_package_decl` what you meant to rename here, not `com_en_declaration`?

plugin/ensime.vim
jvican: `@execute_with_client()` was introduced explicitly to remove the incredible boilerplate code that was available before and `(args, range)` allows us to generalize over all the methods and gives us the ability to customize the external API of a method in an easy way. AFAIK, it is a popular pattern used throughout all the vim plugins and community. It was present in the plugin even before I started to contribute.
  
Regarding the issue that you point out about `autoload` and the wrappers, could you please elaborate on what you mean and what are you proposing? Good ideas are welcome :+1: 
  
And yes, `com_` comes from 'command'.

jvican:
If you change the typo that @ches catched, LGTM.

ChrisCoffey:
Thanks for the review guys. I'm happy to have the conversation on Gitter or the issue if you decide to make one.

ChrisCoffey:
Fixed. Please take another look and merge if you're comfortable.

plugin/ensime.vim
ches: > Regarding the issue that you point out about autoload and the wrappers, could you please elaborate on what you mean and what are you proposing? Good ideas are welcome 👍 

Basically I meant, do we need the `function! EnPackageDecl()` at all? All that it does is call the `ensime#fun_en_package_decl()` function, so couldn't we use the latter function directly and not have two pieces of code to maintain? Unless they are script-scoped with `s:`, all of the functions are "public API" already, they can be called from user code/config. It's best for us to think of them this way when designing the plugin IMO, because it is a fact we can't avoid 😃 

We don't have a ton of functions on the VimL side yet except for those that dispatch for commands, since substantial implementation is on the Python side, but I'm sure the Vim API will grow over time as users want it.

ches: Okay this comment is going back into the weeds of off-topic for the pull request, but I'm not sure where to put it yet other than in this conversation context, so I'm just leaving it here and can refer to it later from wherever a better place is…

> `(args, range)` allows us to generalize over all the methods and gives us the ability to customize the external API of a method in an easy way. AFAIK, it is a popular pattern used throughout all the vim plugins and community.

I can sort of understand reasoning for this *if it's for functions that implement commands specifically* (the `com_` ones in our case), but if it starts to proliferate to everything else—as it is in this pull request—I feel it's a bad habit. It's like having a function interface that's mostly meaningless, you can't tell at all how anything should be called unless there is documentation of what the parameters should hold (we have none so far), or else you have to read the code every time to see how they're used within each function. That gets tiresome. We're choosing to make refactoring easier for we the developers, _maybe_, at the expense of poor API for users (at least users who are sophisticated enough to explore the functions that are exposed, not just the documented commands).

I can cite lots of well-written Vim plugins that use meaningful function parameters:

- [Unite](https://github.com/Shougo/unite.vim/blob/master/autoload/unite.vim)
- [lexima](https://github.com/cohama/lexima.vim/blob/master/autoload/lexima.vim)
- [vim-go](https://github.com/fatih/vim-go/blob/master/autoload/go/util.vim)

To name a few that are instructive to read.

> It was present in the plugin even before I started to contribute.

I don't mean to be critical of anyone personally, we have code that works and is useful today, and it got there through your organic improvements. I'm just sharing some thoughts where I've felt there is room for continuing refinement, from working with the code a little bit so far.

It would be better to make some examples via a PR and discuss from there, and I'll try when I have the time, but it's a case where it could end up being a big, out-of-nowhere refactoring PR that wastes a lot of time if we didn't have some discussion beforehand ☺️

ches:
Summarizing a few directly relevant thoughts extracted from the wider discussion, and a new one or two:

- I'm not sure there's any need for the `EnPackageDecl()` wrapper function?
- Could we drop the `fun_` from the names, on the Vim-side functions at least? I can see how the sort-of-Hungarian-style naming can be a helpful signal when it's the implementation of a command, but when it's a function… it _is_ a function. Seems redundant.
- Since it's not for a command, can the `(args, range)` params be removed since they have no meaning? I'm not sure if that makes `@execute_with_client()` unusable, I'm just asking.

ChrisCoffey:
@ches I'm quite interested in removing the `fun_` prefix, that's always bothered me. I've created #243 & #244 to track your points & I may as well remove the `EnPackageDecl()` wrapper in this pull.

plugin/ensime.vim
jvican: There's been a misunderstanding, I thought you were talking about the abstraction of `(args, range)` introduced in the wrappers of python as keywords, not about this viml code. In that case, I'm sorry, I completely agree this is something we should remove. Also, I do know this is not something personal. This is programming, let's keep egos away.
  
Are you up for a quick PR to remove those methods and make those renamings? Since you're the one that proposed them, I think it's fair if you carry them out. I haven't read those vim plugins, but I'm sure I've seen this pattern elsewhere, and in popular vim plugins. I do not know if it's a good practice in the community, though.
  
I'm waiting for your PR, cheers!

jvican:
Let me know if you stumble upon some problems with `@execute_with_client`.

ChrisCoffey:
@jvican I'd rather not add to this pull if I can help it. This was meant to address current issues with the workflow. I'd prefer to address the refactoring separately. 

In light of that, assuming you're okay with it, is this good to merge?

plugin/ensime.vim
ches: > Are you up for a quick PR to remove those methods and make those renamings? Since you're the one that proposed them, I think it's fair if you carry them out.

Sure, it seems that it's been on my mind already so I ought to act on it 😇  Just trying to find the time… The renaming part should be quick but other aspects might lead deeper. I'll start with #243.

jvican:
This is good to merge.

Issue #246, opened 2016-04-29
Refactor EnsimeClient into Mixins


ChrisCoffey:
The `EnsimeClient` has a lot going on, and I personally have found it a bit painful to navigate. I'd like to factor out the Package Inspector, Debugger, & Symbol Browser components into their own mixins. 

This should only be a code re-org, I don't intend to change any logic. That being said, my hope is that this may make unit testing a bit simpler as well (at least insofar as we can easily focus on a single module).

jvican:
W.r.t. the test suite, I agree that this will make it easier. However, I'd expect to have a minimal PR of the test suite before starting to tackle this issue. At the point when there's a a minimal PR, I think it'd be a good idea to evolve both PRs jointly. I'd propose this to ensure spending time on this when it's not meaningful for the test suite, which should have highest priority.
  
Also, just a heads-up, `EnsimeClient` is very stateful so be careful when you refactor it :smiley:.

ChrisCoffey:
I can see the test first then cleanup viewpoint. I was finding myself spending more time jumping through the main file than I wanted though, hence the mixin class. 

Rather than evolving PR's I'd prefer small incremental changes to land before the next PR does. I'm happy to chat about this more on Gitter if you want.

ches:
Chipping away at the monolith that is `EnsimeClient` has been happening bit by bit as a matter of course. #325 acknowledges the monolith and has some WIP nearly finished, #328 moved lots of Vim stuff to the `Editor` class already, and #366 presents a fuzzy vision for *really* dismantling it.

There is a unit test suite now and mixins as code re-org alone haven't been a great help to testability since they haven't removed `EnsimeClient`'s coupling to complex dependencies—what I referred to as "pushing food around on the plate" in #366. I think we're finally reaching a breakthrough point for testing with the `vim` object and `EnsimeLauncher` both being removed from the client.

So, I think there's nothing left for this ticket to stand for now—the gears are in motion 😄

PR #247, opened 2016-04-29
Move Debugger to Mixin to make future Debugger Extension simpler

ChrisCoffey:
I've found working with the single giant `EnsimeClient` class a bit cumbersome & would like to start factoring slices of the Ensime API out into mixins. I started working on the Debugger this morning, so this was a logical first step.

#246 tracks this idea.

### DIFF
--- a/ensime_shared/debugger.py
+++ b/ensime_shared/debugger.py
@@ -1,0 +1,63 @@
+from ensime_shared.config import feedback
+
+class DebuggerClient(object):
+    """This is the implementation of the Ensime debugger client, it must be mixed in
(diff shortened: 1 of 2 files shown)


jvican:
Despite my comments in @246, this LGTM. Is it ready to merge?
  
As I commented before, I'd suggest to put aside the other refactorings since they're not as meaningful as the debugging layer and there's no real benefit from them (until we have a minimal test suite).

ChrisCoffey:
Yes, this is ready to merge & I agree. Agreed on the other refactorings, #246 is meant to track this but it's not at the top of my priority list.

PR #250, opened 2016-05-01
Fix EnClasspath when running in fish shell

### LINKED ISSUE #249: Hangs on classpath generation
Running `:EnClasspath` hangs with the message `Waiting for generation of classpath...`
on `NVIM 0.1.4 after #242

Exploring `top` shows that sbt was not started, so I believe there is a problem with the terminal command constructed.

The issue was fixed by reverting 
`self.vim.command("terminal sh \"({} && {})\"".format(cd_cmd, sbt_cmd))`
to
`self.vim.command("terminal ({} && {})".format(cd_cmd, sbt_cmd))` -- old version
or 
`self.vim.command("terminal sh \"$({} && {})\"".format(cd_cmd, sbt_cmd))`

I think the underlying problem is `(cmd)` works as command substitution for fish so `cmd` is passed on to sh, but for other shells command substitution is in the form `$()`

The only workaround I can think of is detecting the shell in vim and constructing the command accordingly

yyadavalli:
Fixes #249. Need to use different sublist separation and exit status variables when running in a fish shell

### DIFF
--- a/ensime_shared/launcher.py
+++ b/ensime_shared/launcher.py
@@ -63,7 +63,7 @@ class EnsimeLauncher(object):
         self.vim = vim
         self.base_dir = os.path.abspath(base_dir)
         self.ensime_version = "0.9.10-SNAPSHOT"
-        self.sbt_version = "0.13.8"
+        self.sbt_version = "0.13.11"
 
     def launch(self, conf_path):
         conf = self.parse_conf(conf_path)
--- a/ensime_shared/launcher.py
+++ b/ensime_shared/launcher.py
@@ -142,11 +142,17 @@ class EnsimeLauncher(object):
         inside_nvim = int(self.vim.eval("has('nvim')"))
         if inside_nvim:
             import tempfile
+            import re
             tmp_dir = tempfile.gettempdir()
             flag_file = "{}/ensime-vim-classpath.flag".format(tmp_dir)
             sbt_cmd += "; echo $? > {}".format(flag_file)
             self.vim.command("echo 'Waiting for generation of classpath...'")
-            self.vim.command("terminal sh \"({} && {})\"".format(cd_cmd, sbt_cmd))
+            if re.match(".+fish$", self.vim.eval("&shell")):
+                sbt_cmd += "; echo $status > {}".format(flag_file)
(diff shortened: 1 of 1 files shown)


ensime_shared/launcher.py
```diff
@@ -142,11 +142,17 @@ def generate_classpath(self, scala_version, classpath_file):
...
         if inside_nvim:
             import tempfile
+            import re
             tmp_dir = tempfile.gettempdir()
             flag_file = "{}/ensime-vim-classpath.flag".format(tmp_dir)
             sbt_cmd += "; echo $? > {}".format(flag_file)
             self.vim.command("echo 'Waiting for generation of classpath...'")
-            self.vim.command("terminal sh \"({} && {})\"".format(cd_cmd, sbt_cmd))
+            if re.match(".+fish$", self.vim.eval("&shell")):
+                sbt_cmd += "; echo $status > {}".format(flag_file)
```
jvican: There's some room for improvement here.
```python
target = "$status" if re.match(".+fish$", self.vim.eval("&shell")) else "$?"
sbt_cmd += "; echo {} > {}".format(target, flag_file)
self.vim.command("terminal ({} && {})".format(cd_cmd, sbt_cmd))
```
Why not this?

jvican:
Thanks for the PR. Waiting for your changes/comments! :+1:

ensime_shared/launcher.py
yyadavalli: The command is also different for both shells
`self.vim.command("terminal {}; and {}".format(cd_cmd, sbt_cmd))`
vs
`self.vim.command("terminal ({} && {})".format(cd_cmd, sbt_cmd))`

`fish` does not accept `&&` and cannot handle parentheses `(cmd)`

jvican: Okey, sorry, didn't notice. Is line 148 repeated though?

yyadavalli: @jvican Yup, cleaned it up, thanks.

jvican:
Cool, LGTM! Since this is critical, let's merge it right now. (/cc @ChrisCoffey)

Issue #232, opened 2016-04-08
Standardize Commands to use Arguments instead of Input Prompts


ChrisCoffey:
There are two issues here:

1. We use arguments for some commands (`:EnShowPackage`), but prompt users for input on others (`:EnSearch`). This behavior is confusing for pretty much everyone, and the prompt behavior is less expected.
1. In addition to being somewhat unexpected, prompts are a bit more difficult to script against as they require the user to remember to add a `<cr>` to trigger the prompt before sending their input. This is pretty unusual in Vim and something we should shy away from.

ches:
It's cross-referenced above but just so it's obvious, see prior discussion on #229.

PR #251, opened 2016-05-02
Convert SymbolSearch to use arguments instead of prompt

ChrisCoffey:
This addresses #232 . All other commands already support arguments.


jvican:
Cool, LGTM!

Issue #233, opened 2016-04-08
Use current file/scope's package by default for :EnShowPackage


ches:
This is an enhancement idea, see #229 for some history.

`:EnShowPackage` currently requires typing out a fully qualified name as an argument to the command, which is kind of painful. It's often (usually?) the case that you want to use it for the package you're currently editing, so we could try to make that work when you invoke the command with no argument.

ches:
This was closed by #252. I believe there are still edge cases remaining like nested packages, so I created #286 to keep track of that for the future.

PR #252, opened 2016-05-02
Add default for Package Inspector

ChrisCoffey:
This is an imperfect parser that extracts the current package name from a Scala or Java file, as well as the plumbing to use it for the package inspector.

Previously, we were requiring users to type the FQN they wanted to explore. Now simply running `EnShowPackage` will explore the current package. Hopefully this is a big improvement in usability.

Addresses #233

### DIFF
--- a/ensime_shared/util.py
+++ b/ensime_shared/util.py
@@ -21,6 +21,22 @@ class Util:
         if not os.path.exists(path):
             os.makedirs(path)
 
+    @staticmethod
+    def extract_packageName(lines):
+        foundPacakge = False
+        package = ""
+
+        for line in lines:
+            if "package" not in line and not foundPacakge:
+                continue
+            elif "package" in line:
+                if not package:
+                    package = line.split("package ")[-1].replace("\n", "")
+                else:
+                    package += "." + line.split("package ")[-1].replace("\n", "")
+            else:
+                break
+        return package
 
 @contextmanager
 def catch(exception, handler=lambda e: None):
@@ -37,3 +53,4 @@ def module_exists(module_name):
         __import__(module_name)
         res = True
     return res
+
--- a/ensime_shared/util.py
+++ b/ensime_shared/util.py
@@ -22,7 +22,7 @@ class Util:
             os.makedirs(path)
 
     @staticmethod
-    def extract_packageName(lines):
+    def extract_package_name(lines):
         foundPacakge = False
         package = ""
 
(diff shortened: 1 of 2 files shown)


ensime_shared/util.py
```diff
@@ -21,6 +21,22 @@ def mkdir_p(path):
         if not os.path.exists(path):
             os.makedirs(path)
 
+    @staticmethod
+    def extract_packageName(lines):
+        foundPacakge = False
```
jvican: Minor style issue: don't use camel case in python, use underscores to separate names.

ChrisCoffey: Good catch, thanks

yyadavalli:
Please correct if I'm wrong, this parser works for package declaration at file level like `package alpah.beta` but not for scoped packages

```
package alpha {

   package beta {
       // code or currrent cursor
   }
}
```
One way to find such packages would be to parse upwards from current cursor position for the first occurrence of `package` and get the name from there.
Again this may be unnecessarily complicating a nice and simple implementation

ChrisCoffey:
Yup, this falls down on a handful of edge cases, specifically the package nesting or the term `package` present in comments above the actual package declaration. 

Handing the nesting wouldn't be too bad, but you're correct that having to match braces requires a larger implementation.

Personally I'm okay not providing support for nested packages with this parser. If it turns out this is important to the community I'm more than happy to add it or pull in a library that already does it.

ensime_shared/util.py
```diff
@@ -22,7 +22,7 @@ def mkdir_p(path):
             os.makedirs(path)
 
     @staticmethod
-    def extract_packageName(lines):
+    def extract_package_name(lines):
         foundPacakge = False
```
jvican: There's camel case here as well :smile:

jvican:
Nice work @yyadavalli. I think we should merge this as it is, but your corner cases are very valuable. We should definitely cover them. Would you mind @ChrisCoffey implementing them in another PR? Otherwise, if you want to pick it @yyadavalli, it's all yours :dancer:

yyadavalli:
@jvican  we should definitely merge this and work on improvements later, I imagine this already covers majority of use cases

jvican:
Yes, what I wondered is if you would like to improve it, since you've been the one that spotted it.
  
Merging this, LGTM.

jvican:
No, sorry, forgot that @ChrisCoffey has still to make one renaming. Then, we'll merge.

yyadavalli:
@jvican definitely, if @ChrisCoffey doesn't want to spend more time on it, I will do it

PR #259, opened 2016-05-07
Avoid moving cursor in `complete_func`. Closes #238

### LINKED ISSUE #238: Create a deoplete source
It would be great if this ensime had a deoplete source, documented [here](https://github.com/Shougo/deoplete.nvim/blob/master/doc/deoplete.txt#L541), for asynchronous code completion.

ktonga:
It's not needed when called from __Vim__ (tested
with `Ctrl-x Ctrl-o` and double checked Vim's documentation,
it's not in `complete-functions` specifications), but it breaks
omnicompletion when called from __deoplete__.  

With this fix Ensime completion can be configured as
`g:deoplete#omni#input_patterns` instead of
`g:deoplete#omni_patterns`. See 'FAQ' in __deoplete__
documentation to know more about the advantages.  

Sample config:  
```VimL
let g:deoplete#omni#input_patterns = {}
let g:deoplete#omni#input_patterns.scala = '[^. *\t]\.\w*'
```

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -1099,7 +1099,6 @@ class EnsimeClient(DebuggerClient, object):
             result = []
             # Only handle snd invocation if fst has already been done
             if self.completion_started:
-                self.vim_command("until_first_char_word")
                 # Unqueing messages until we get suggestions
                 self.unqueue(timeout=self.completion_timeout, should_wait=True)
                 suggestions = self.suggestions or []


jvican:
Please, before merging this, could you double-check that it works in neovim?

ktonga:
Hi @jvican I tested it in Neovim. My main motivation for fixing this is to be able to use ensime-vim completion with deoplete which is a neovim-only plugin.

jvican:
Sorry, totally miss deoplete was neovim-only.
  
LGTM. This is all yours @ChrisCoffey.

Jacoby6000:
@ChrisCoffey I know this is already closed, but would you be against contributions that automatically configure deoplete?

jvican:
WDYM @Jacoby6000 ?

Jacoby6000:
Would it be reasonable for ensime-vim to provide deoplete configuration?  Then, rather than people having to work out what completion patterns should be, it's just supported by ensime.

jvican:
Yes, it would be reasonable! I guess that they should be pretty standard and improve the user experience. :smile:

PR #264, opened 2016-05-17
Consitently place the log file in .ensime_cache

yyadavalli:
We are placing the log file in `tmp` if `.ensime_cache` doesn't exists. This changes that and tries to create `.ensime_cache` folder.

Please advice if I should handle the case where folder creation fails


jvican:
Yes, please handle that case and fallback to tmp folder if it fails.

ChrisCoffey:
this LGTM. Thanks for all the pr's lately @yyadavalli

Issue #236, opened 2016-04-19
Prevent the launcher from being started if user hasn't installed sexpdata


ches:
See the unfinished TODO from #235, this seems to be catching people off guard so it ought to get done soon. I'll try to find time.

~~I'm also about to send a PR for the doc site to try to make it more obvious in the installation instructions.~~

Half of what I originally intended this to be, @ChrisCoffey took care of in #265—it was easier than I thought it was on first look. The other half I believe was that the launcher can still be started if `sexpdata` is missing, and that will result in an exception. This should be prevented. The path to launcher invocation is awfully convoluted too, it would generally benefit from refactoring to simplify it.
#294 is also related, possibly could be a prerequisite if it could be used to avoid this problem.

ensime-vim should be disabled if this or other critical deps are missing

jvican:
:+1:

PR #265, opened 2016-05-18
Warn at startup if sexp package cannot be found

ChrisCoffey:
Pretty small change using the existing warning code from websockets.

Addresses #236


jvican:
LGTM.

PR #266, opened 2016-05-18
Append jar file at beginning if present

### LINKED ISSUE #260: Support Local Assembly Jar
Right now Emacs supports an alternative startup mode where the presence of a local version of the Ensime server jar circumvents the classpath file. This would be extremely useful for local testing.

Right now my workaround is to bump the server version, publish locally, bump the version the plugin looks for, then remember to roll back all the version changes & remove the local version from my `.ivy` cache. This is a pretty messy procedure that I forget to unwind sometimes.

yyadavalli:
Check if assembly jar matching project's scala version is present and append at beginning of the classpath

Resolve #260

### DIFF
--- a/ensime_shared/launcher.py
+++ b/ensime_shared/launcher.py
@@ -100,7 +100,7 @@ class EnsimeLauncher(object):
             Util.read_file(classpath_file), java_home)
 
         for x in os.listdir(self.base_dir):
-            if fnmatch.fnmatch(x, "ensime_" + scala_version[:4] + "*.jar"):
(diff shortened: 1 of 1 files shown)


yyadavalli:
Please comment if the patter to match should be more exact
 `"ensime_" + scala_version[:4] + "*-assembly.jar"` vs 
`"ensime_" + scala_version[:4] + "*.jar"`

jvican:
Yes, it should be more exact. Otherwise, LGTM.

ChrisCoffey:
Nice work!

PR #278, opened 2016-06-08
fix syntax error

tek:
`except Exception, e:` is python2-only syntax

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -641,7 +641,7 @@ class EnsimeClient(DebuggerClient, object):
                 try:
                     if webbrowser.open(url):
                         self.log("opened {}".format(url))
-                except webbrowser.Error, e:
+                except webbrowser.Error as e:
                     log_msg = "handle_string_response: webbrowser error: {}"
                     self.log(log_msg.format(e))
                     self.raw_message(feedback["manual_doc"].format(url))


yazgoo:
LGTM, I think we'd rather support python versions > 2.5 than versions < 3 :)

http://stackoverflow.com/a/2535770

tek:
though I just noticed that it only worked with py3 here because I renamed `rplugin/python` to `rplugin/python3`. Is there a secret to making this work that I'm not aware of?

ChrisCoffey:
I'm happy to land this based on the syntax change, although I'm curious what you're seeing between `rplugin/python` & `rplugin/python3`

tek:
well, having `python3` as `python_host_prog`, `:UpdateRemotePlugins` fails with `Vim(let):E903: Could not spawn API job`, with the debug log not mentioning anything about it. Unless I rename the rplugin dir to `python3` and fix the path in `ensime_init_path()`, then it just works. Not sure what the right course of action would be here.

jvican:
Could you please test if a symlink from `python3` to `python` would work @tek? I've got the feeling that if we rename `python` to `python3` directly, support for `python2` would vanish.

tek:
if I replace `path.startswith` with a regex match test, it works, but the python2 plugin will still be attempted to be loaded. I guess this is a problem with neovim's plugin host initialization routine, I'll have a closer look at it to see if this can be avoided.
FWIW, for my plugins, I publish the python part on pypi and include only a dummy adapter in the neovim plugin, which makes handling deps and pythonpath trivial.

tek:
alright, it was just a configuration error after all. I had set `python_host_prog = 'python3'`, but that variable is intended for the python2 interpreter. Unsetting it solved the problem. I apologize.

Issue #277, opened 2016-06-07
Aggregate errors before passing them on to syntastic until the FullTypeCheckCompleteEvent is received


atnoya:
When performing big refactorings, syntastic blocking behaviour really gets in the way, it really makes neovim pretty unusable for quite a bit of time when the number of errors is pretty big

Would like to suggest/ask for a way to tell ensime-vim to aggregate the new error notes until the `FullTypeCheckCompleteEvent` is received. Appending all them together upon the reception of such event.

PR #279, opened 2016-06-09
Add Buffered TypeChecking

ChrisCoffey:
This is one approach to address #277 . Please give it a look & let me know what you think. It's a pretty stateful workflow, so I don't feel too bad about using globals.

I need to test this more, I'm mostly putting this up as a strawman. If people like it I'll tighten up the code.


jvican:
I like it. Could you add more documentation to both methods and variables? It's handling a rather special issue that needs more explanation for future contributors.

atnoya:
I like it, I would love to see this in place 😄  Thx for implement my suggestion!

plugin/ensime.vim
```diff
@@ -12,6 +12,7 @@ augroup END
 
 command! -nargs=* -range EnNoTeardown call ensime#com_en_no_teardown([<f-args>], '')
 command! -nargs=* -range EnTypeCheck call ensime#com_en_type_check([<f-args>], '')
+command! -nargs=* -range EnBufferedTypeCheck call ensime#com_en_buf_type_check([<f-args>], '')
```
ches: I know this is rough-ish implementation, but just asking for sake of design discussion: does this really need to exist as a separate command?

If a bit of testing proves this to be robust and trouble-free, it seems entirely reasonable to me for this to be in the default code path for `:EnTypeCheck`. If it's potentially controversial, then *maybe* have an option to disable the buffering, but I lean toward not having tons of options until/unless someone makes a good case for wanting them.

jvican: Good point, I would tend to agree. It could start as an option that you can toggle and then become the only supported `typecheck`.

ChrisCoffey: I agree @ches, adding this as the default style of typechecking w/ an option to disable if people want it is a nice simplicity win. I'll clean this up, remove the command & make it the default approach & then bang away on it for a while.

ChrisCoffey:
This works great in my local testing, but I'd love if you guys would apply it as a patch and bang on it for a bit or suggest some edge cases for me. 

Otherwise, I'm happy with this as the default; I think it makes the code much simpler & the user experience better. Thanks for the suggestion @ches

Jacoby6000:
Appears to be working for me.  My only complaint is how long it takes to get feedback from the typechecker.  Is there a way to add a message that says "Type Checking..." in the status bar or something?

rorygraves:
@Jacoby6000 Does the file you are checking have lots of type magic?  The limitation may be the presentation compiler.

Jacoby6000:
I don't think so. The file I'm in was a simple Test class. The most type magic going on was probably the matchers. I'm fine with it taking awhile to respond, but some indicator that it's working would be nice.

ChrisCoffey:
Sure, I can add a message

ChrisCoffey:
I'm going to add some tests around this code before it lands.

ensime_shared/config.py
```diff
@@ -8,6 +8,7 @@
     "unknown_symbol": "Symbol not found",
     "indexer_ready": "Ensime indexer is ready",
     "analyzer_ready": "Ensime analyzer is ready",
+    "typechecking": "Typechcking...",
```
ktonga: Typo

ktonga:
Big 👍 to the factor out.

ensime_shared/vim_commands.py
```diff
@@ -0,0 +1,44 @@
+commands = {
```
jvican: Not sure this is a great idea. People usually want to have this in the same file where all the logic is. Also, it allows future developers to realise that they shouldn't embed any vim command in the code, but factor it out.

ChrisCoffey: Given the size of the `ensime.py` file I'm not sure how much of a difference it really makes. It's easy to notice that other functions are looking something up from the commands dictionary & that should be the pattern followed.

That being said, if you feel strongly against factoring this out I will put it back.

ches: Just noticed, FWIW, the same thing is already basically done with the `feedback` dict. If we keep the extraction of `commands`, maybe it could at least go in `config.py` alongside `feedback` just to reduce proliferation of files.

jvican: Yeah, let's put it in the `config.py` file and that's it. I don't feel strongly about this.

PR #281, opened 2016-06-10
Fix dev env

ktonga:
- Lettuce was being run without virtualenv
- Setup venv to use Py2 since Lettuce doesn't support Py3

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,7 +2,6 @@ build:
   image: ensime/ensime:latest
   pull: true
   commands:
-    - pip install -r requirements.txt
-    - pip install -r test-requirements.txt
     - pip install python-coveralls
-    - lettuce ensime_shared/spec/features
+    - make run-tests
+
--- a/.drone.yml
+++ b/.drone.yml
@@ -2,6 +2,7 @@ build:
   image: ensime/ensime:latest
   pull: true
   commands:
+    - pip install virtualenv
     - pip install python-coveralls
     - make run-tests
 
--- a/.drone.yml
+++ b/.drone.yml
@@ -5,4 +5,4 @@ build:
     - pip install -r requirements.txt
     - pip install -r test-requirements.txt
     - pip install python-coveralls
-    - make run-tests
+    - lettuce ensime_shared/spec/features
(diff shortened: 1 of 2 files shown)


.drone.yml
```diff
@@ -5,4 +5,4 @@ build:
     - pip install -r requirements.txt
     - pip install -r test-requirements.txt
     - pip install python-coveralls
-    - make run-tests
+    - lettuce ensime_shared/spec/features
```
ches: Rather than change this from using the `make` target and leaving to two places where the test procedure has to be updated and maintained, it would make more sense to me to make a few adjustments to the `Makefile` and otherwise leave the CI config running `make run-tests`:

1. Make the `run-tests` target depend on `venv` one.
2. Make the `venv` target additionally depend on `$(ACTIVATE)`, so that it triggers creation of the virtualenv if it doesn't exist yet.
3. Remove the `pip install -r *.txt` steps here in the CI config as they're made redundant.

I think that'll do it, but haven't actually tried yet so I might be overlooking something.

ktonga: I agree. It was the easiest way to fix it, but I will change it to don't DRY.

ktonga: @ches is `venv` idempotent? I don't feel good running it every time I want to run the tests.

ches: I think, from reading without actually trying to change it yet, that it would be idempotent and not wastefully run `pip install` when not needed, as long as `$(ACTIVATE)` exists and `requirements.txt` hasn't changed, since it includes the `test -d .venv` check. IMO it could be refactored a little bit to be more clear and let make do the work, maybe sort of like this (untested):

```make
# Moved to first task to be the implicit default
run-tests: requirements
	# ...

$(ACTIVATE):
	virtualenv -p python2 .venv

requirements: $(ACTIVATE) requirements.txt
	. $(ACTIVATE); $(REQS)
	. $(ACTIVATE); $(TESTREQS)

lint: requirements
	# ...

autopep8: requirements
	# ...
```

Also some/most of the targets technically should be declared as `.PHONY`. The newly-working-again test suite was just merged, I've never really looked at the `Makefile` until now 😇

ktonga:
Done. `make run-tests` triggers __virtualenv__ setup if needed and only runs the tests if `.venv` is up-to-date. It works locally and in drone.

jvican:
LGTM

PR #285, opened 2016-06-17
Fix missing function shims for standard Vim

ches:
Pretty straightforward, a couple of function shims were missing on the standard Vim side, so the `:EnAddImport` and `:EnOrganizeImports` commands weren't working.


jvican:
LGTM :+1:

PR #287, opened 2016-06-17
Make help doc comprehensive and fully tagged; little meta bits

ches:
So I sat down to get caught up with the project, and this seemed like a suitable time to do some documentation. I believe I've pretty thoroughly covered the capabilities and state of the project from a user perspective here. I still feel that the docs on the web could use some work, so I tried to think with that in mind as well—some of this, like the Getting Started section maybe, could be moved or copied to the web to supplement the current docs. Let me know what you think about that and I can put together a PR for the docs repo.

Vimdoc is a loosely-structured format so there's always a degree of aesthetic preference involved. This is just a style I've arrived at over time. If you ever care to edit at length, I have some UltiSnips helpers [here][1] extending the ones [here][2].

There are a couple of other unrelated changes here, they're not functional changes (well okay, a couple lines of VimL include guard) so it was convenient to just put them on the same branch. But they're separate commits so if you find this distracting I can split them out. Rationale for `Makefile` updates is well-described in the commit message.

[1]: https://github.com/ches/dotfiles/blob/3200b58f028403f99a03c216a95f127fff2cfcc6/.vim/UltiSnips/help.snippets
[2]: https://github.com/honza/vim-snippets/blob/1d9d0413fc77bc1bb893e988f68e19f9b0bdb97b/UltiSnips/help.snippets


ches:
Doh, I added the optional `python-Levenshtein` test dep that Lettuce barks about being a speed up, but looks like it needs something probably like a `python-devel` dependency on the drone.io environment. I'll look into this in a few hours, I have to step offline.

doc/ensime.txt
```diff
@@ -0,0 +1,509 @@
+*ensime.txt*	A client for ENSIME, IDE-like features for Scala and Java.
+
+Author:  Olivier 'yazgoo' Abdesselam
+         Jorge Vicente Cantero (jvican)
```
jvican: It would be great to add here @ChrisCoffey!

jvican:
Documentation is thorough and clear to read, thanks for the PR :+1:. LGTM.

doc/ensime.txt
ches: I left that decision to your authority, unless you want me to add it now 😉

ches:
I've opened a PR to get the Python devel headers added to the CI container image so that `python-Levenshtein` should build (in the references above). Assuming that's merged soon, we can re-trigger a build on this, or I can strip the commit with the build changes off this branch for now as I probably should have just done in the first place 😓 

@ChrisCoffey how's this look to you assuming the build is green soon?

doc/ensime.txt
jvican: Yeah, let's add him to the list now ;)

ches:
Okay so Python development headers are merged for the Docker image but actually building updates to it has some… complications right now. But I guess we learned something useful there, seems we ought to be using the `v1.x-cache` Docker tag. Now done here.

Commented out the optional `python-Levenshtein` test requirement for now, squashed a couple of noisy clean-up commits—build is green.

jvican:
LGTM.

ChrisCoffey:
This is excellent! Thanks @ches

Issue #276, opened 2016-06-06
Add type params list to method signature in completion popup


ktonga:
Depends on ensime/ensime-server/#1504

ches:
Looks like the server side is coming along nicely in ensime/ensime-server#1539 / ensime/ensime-server#1546, nice work 👏

ktonga:
We will need to support server 2.0 to add this

ktonga:
Development of this task can be started after solving #300

Issue #273, opened 2016-06-03
Completion: Several problems regarding type formatting


ktonga:
In the completion popup, specially for __functions__, the type hint may be ok for simple signatures, but it easily gets messy for complex ones. Here you have an example for `Seq.fill`:

What __Intellij Idea__ reports:
<img width="709" alt="screen shot 2016-06-03 at 9 59 43 pm" src="https://cloud.githubusercontent.com/assets/1324947/15777970/907c371a-29d6-11e6-8507-a0c4ad3606ca.png">

What __ensime-vim__ reports:
<img width="1384" alt="screen shot 2016-06-03 at 9 18 01 pm" src="https://cloud.githubusercontent.com/assets/1324947/15777613/1e5b03b6-29d4-11e6-9883-7b3238f4e428.png">

I can see a few problems so far:

- [x] Only the first entry is shown for overloaded functions (but all are present in server's response)
- [ ] Type parameters are not included at all
- [x] Only the first argument list is used
- [x] Fully qualified names are used which gets pretty unreadable

Now that ensime/ensime-server#1468 has been merged these problems seems quite simple to solve. I will start working on it and hopefully I'll be raising a PR in a few days.

ChrisCoffey:
Sounds great. Please ask if you get stuck.

jvican:
Regarding the "fully qualified names" issue, you can work around it by using the ensime command `EnToggleFullType`.

ktonga:
@jvican I checked the code and it formats the signature directly using the "sections" field which only has the full names and is hard to manipulate due to its lack of structure.

On a side note, I don't think it's necessary to make it configurable. 93% of the times, full names in the signatures is useless and confusing. I saw this number once in a Harvard research :P

ktonga:
Well, today I was able to make some progress, and this is the result so far:

Before
![snapshot12](https://cloud.githubusercontent.com/assets/1324947/15830185/5499b84a-2c5a-11e6-8d68-ee248e4895c8.png)

After
![snapshot13](https://cloud.githubusercontent.com/assets/1324947/15830200/656c70b8-2c5a-11e6-8c8a-b6fd0da70ed1.png)

It's almost finished, I'm only missing a way to add the type params to the generic methods, but I cannot figure it out with the actual structure of `typeInfo`. I've already asked about it in __ensime-server__ gitter channel.

I have a few more improvements in mind but I'll propose them in separate issues/PRs.

Cheers.

ktonga:
[...]

PR #288, opened 2016-06-21
Make qualified vs. shallow types mode actually work consistently on type inspections

ches:
I started out innocently just wanting to add echo feedback when you toggled `:EnToggleFullType` on and off. Then I realized the support for this feature didn't actually work very well.

`:EnType` ignored it completely, always showing qualified types, and that's the command where I most often wanted it disabled. `:EnType` had some other problems too—before this branch, with this (silly) code:

```scala
import scala.concurrent.Future

val fut = Future.successful("bippy")

def carTrip(times: Int, msg: String) = for (_ <- 0 to count) yield msg
carTrip(5, "Are we there yet?")
```

`:EnType` on `fut` would give:

```scala
scala.concurrent.Future[java.lang.String][java.lang.String]
```

Note the duplicate type parameters. For a function type, do `:EnType` on `carTrip` and you'd get this:

```scala
(_0: Int, _1: String) => scala.collection.immutable.IndexedSeq[java.lang.String]
```

Shallow types for the params, qualified for result, and those parameter names are noisy and pretty useless. These will now look like this:

```scala
// The default with shallow types
Future[String]
(Int, String) => IndexedSeq[String]

// With qualified toggled on:
scala.concurrent.Future[java.lang.String]
(scala.Int, java.lang.String) => scala.collection.immutable.IndexedSeq[java.lang.String]
```

Similar changes have been applied to `:EnInspectType`. **Side note**: I think the formatting of using parens and `=>` for the parent types on `:EnInspectType` is suboptimal, it looks like a function type and that's even harder to grok when you actually inspect a function type. I haven't yet thought of a good alternative presentation though, suggestions welcome.

I'm going to think about tests for this, some refactoring would make tests easier to write but that would also make review harder. So I went ahead and pushed what's here for now.

### DIFF
--- a/ensime_shared/config.py
+++ b/ensime_shared/config.py
@@ -8,8 +8,8 @@ feedback = {
     "displayed_type": "The type {} has been displayed",
     "failed_refactoring":
         "The refactoring could not be applied (more info at logs)",
-    "fulltype_display_on": "[ensime] Qualified type display enabled",
-    "fulltype_display_off": "[ensime] Qualified type display disabled",
+    "full_types_enabled_on": "[ensime] Qualified type display enabled",
+    "full_types_enabled_off": "[ensime] Qualified type display disabled",
     "indexer_ready": "Ensime indexer is ready",
     "manual_doc": "Go to {}",
     "missing_debug_class": "You must specify a class to debug",
--- a/ensime_shared/config.py
+++ b/ensime_shared/config.py
@@ -4,21 +4,24 @@ gconfig = {
 }
 
 feedback = {
-    "start_message": "Ensime server has been started...",
-    "unknown_symbol": "Symbol not found",
-    "indexer_ready": "Ensime indexer is ready",
     "analyzer_ready": "Ensime analyzer is ready",
-    "typechecking": "Typechcking...",
-    "module_missing": "{} missing: do a `pip install {}` and restart vim",
-    "warn_classpath": "Execute :EnClasspath to set a classpath",
-    "missing_debug_class": "You must specify a class to debug",
-    "notify_break": "Execution breaked at {} {}",
     "displayed_type": "The type {} has been displayed",
     "failed_refactoring":
         "The refactoring could not be applied (more info at logs)",
-    "unhandled_response": "Response {} has not been handled",
+    "fulltype_display_on": "[ensime] Qualified type display enabled",
+    "fulltype_display_off": "[ensime] Qualified type display disabled",
+    "indexer_ready": "Ensime indexer is ready",
+    "manual_doc": "Go to {}",
+    "missing_debug_class": "You must specify a class to debug",
+    "module_missing": "{} missing: do a `pip install {}` and restart vim",
+    "notify_break": "Execution breaked at {} {}",
     "spawned_browser": "Opened tab {}",
-    "manual_doc": "Go to {}"}
+    "start_message": "Ensime server has been started...",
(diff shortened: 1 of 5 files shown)


ensime_shared/config.py
```diff
@@ -4,17 +4,20 @@
...
+    "missing_debug_class": "You must specify a class to debug",
+    "module_missing": "{} missing: do a `pip install {}` and restart vim",
+    "notify_break": "Execution breaked at {} {}",
     "spawned_browser": "Opened tab {}",
-    "manual_doc": "Go to {}"}
+    "start_message": "Ensime server has been started...",
+    "unhandled_response": "Response {} has not been handled",
+    "unknown_symbol": "Symbol not found",
+    "warn_classpath": "Execute :EnClasspath to set a classpath"
+}
```
ches: Sorry about the noise here, all that's changed is adding the `full_types_enabled_*` messages but I went ahead and sorted these lines before it keeps growing…

ensime_shared/spec/features/suggestion_format.feature
```diff
@@ -13,22 +13,6 @@ Feature: Format a Completion Suggestion
...
-    Then We get the format [A]
-
-  Scenario: Format multiple type parameters
-    Given A list of type parameters:
-      | tparam |
-      | A      |
-      | B      |
-      | C      |
-    When We concat the type parameters
-    Then We get the format [A, B, C]
```
ches: To be clear, these scenarios actually had nothing to do with the completions menu at all—there is no integration in these tests, they are basically just unit tests for the `concat_tparams` function which is now completely unneeded.

ktonga: I was planning to add type parameters to signatures in completion popup but it needs some extra work in the server. I've created #276 to track it. No worries, I can bring this function back if needed.

ches: I had looked back at #273 and seen that, but I didn't completely understand what you meant on the ensime-server ticket about what was missing. Now I do, after looking at the server responses more closely in that situation, and I agree that type params should be added to the response (and would add minimally to payloads).

As of this moment on this branch, it's dead code and I prefer to remove dead code as soon as it's dead, that's what version control is for 😁  I should have mentioned you or commented on #276, but now you're on top of it already 😄

Jacoby6000:
would it be possible to make `:EnType` return parameter names of functions?  I'll agree that `_0` and `_1` aren't useful. But if the server is providing the parameter names (I don't know if it is.), it'd be nice to see `(times: Int, message: String) => IndexedSeq[String]`.

ches:
The server does not provide the parameter names, in this context.

On the plus side, this output is a syntactically valid type: you could copy-and-paste it verbatim to be the type annotation for a higher-order function parameter, to `:kind` in the REPL, etc.

ChrisCoffey:
LGTM once you rebase

ches:
Ah, didn't realize some conflicts had crept in, rebased. About more decoupling for tests, I'm trying a bit of a new direction that I'll hopefully share soon…

ChrisCoffey:
LGTM. Assuming you're ready to merge @ches, @jvican feel free to merge when you're ready

ches:
Yep, some work toward testability will come separately. I'd like to land this, I think it's an improvement and it's been working for me without issue.

jvican:
Thanks @ches, this looks nice.

Issue #280, opened 2016-06-09
Errors after running :EnClasspath


chipsenkbeil:
So, I'm trying to set up vim with ensime to work on the ensime-server project. From reading the docs online, my assumption was that I do the following:

1. Run `sbt ensimeConfig` from the root of the project
2. Start vim
3. Run `:EnClasspath` once when first starting the project

After running `:EnClasspath`, I'm told to hit enter to continue after various libraries are resolved and downloaded. This results in me returning to vim and seeing this error:

```
Error detected while processing function ensime#com_en_classpath[1]..<SNR>161_call_plugin:
E605: Exception not caught: <type 'exceptions.OSError'>: [Errno 2] No such file or directory
```

Furthermore, attempting to quit (via `:q`) freezes vim, which no longer responds (I can't even interrupt or suspend it). I have to search for the process and kill it via `kill -9 <proc id>`.

Did I miss a step? Something else going wrong? My JVM is as follows:
```
openjdk version "1.8.0_92"
OpenJDK Runtime Environment (build 1.8.0_92-b14)
OpenJDK 64-Bit Server VM (build 25.92-b14, mixed mode)
```

Consecutive calls to `:EnClasspath` without quitting the vim session results in the same error being printed out (but nothing being resolved, just stays in the vim session).

ChrisCoffey:
@chipsenkbeil is there any additional information in the `.ensime_cache/ensime-vim.log`?

chipsenkbeil:
Removed it, opened vim and ran `:EnClasspath` again, and here is the resulting log:

```
2016-06-09 10:28:29.924735: __init__: in
[...]

PR #293, opened 2016-06-29
Make project config part of EnsimeLauncher's state

ches:
This is several incremental refactoring steps that center around making the parsed `.ensime` part of an `EnsimeLauncher` instance's state, which I think makes logical design sense and you'll see that it reduces a lot of repetitive code and redundant method arguments getting passed around. This makes the class's interface simpler which IMO makes the plugin initialization and classpath bootstrap processes somewhat easier to understand.

There are several independently-coherent commits, it is probably easier to first review them one-by-one.

I'm going to try to add some tests here and thus I might rebase to make sure tests passed before these changes with no unexpected behavior changing along the way. But again, like #288, the code often isn't factored to make testability easy yet, so I might lose steam to finish that right now. Went ahead and opened a PR to make others aware there is some WIP in this area, up to you whether we wait on tests to merge.

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -54,7 +54,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, object):
             self.vim_command("set_updatetime")
             self.vim_command("set_ensime_completion")
             self.vim.command("autocmd FileType package_info nnoremap <buffer> <Space> :call EnPackageDecl()<CR>")
-            self.vim.command("autocmd FileType package_info  setlocal splitright")
+            self.vim.command("autocmd FileType package_info setlocal splitright")
             super(EnsimeClient, self).__init__()
 
         def setup_logger_and_paths():
@@ -67,7 +67,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, object):
             if not osp.isdir(self.ensime_cache):
                 try:
                     os.mkdir(self.ensime_cache)
-                except OSError as exception:
+                except OSError:
                     self.log_dir = "/tmp/"
             self.log_file = os.path.join(self.log_dir, "ensime-vim.log")
             with open(self.log_file, "w") as f:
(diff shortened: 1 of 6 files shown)


ensime_shared/ensime.py
```diff
@@ -195,15 +194,20 @@ def lazy_initialize_ensime():
...
                         self.message("warn_classpath")
                     return False
-                self.ensime = self.launcher.launch(self.config_path)
+
+                try:
+                    self.ensime = self.launcher.launch()
+                except InvalidJavaPathError:
+                    # TODO: also disable plugin/shut down server here
+                    self.message('invalid_java')
+
```
ches: I tried doing `self.disable_plugin()` here, but it isn't really very useful—see #294.

The behavior at this point is that it will display the message prompting the user to press a key to acknowledge. That part is good. It possibly requires a key press multiple times initially, not entirely sure why but I presume it might have to do with connection retry logic causing a few messages to be buffered by Vim at one time. Then the user can quit cleanly and fix the config problem, or go on using Vim without ENSIME while the plugin log and message area continue to be spammed with the error over and over (at least non-intrusively as far as not requiring a key press) and Vim performance pays the price of futile autocommand event handler callbacks continuing to be tried.

As-is this is still a user experience improvement for #280, and I believe these caveats will be resolved by addressing #294 and/or some review of logic that retries `setup()` indefinitely regardless of a retry counter (e.g. `CursorMoved` handler).

ensime_shared/errors.py
```diff
@@ -1,9 +1,18 @@
 import os
 
 
-class Error:
-    """Represents an Error in the ensime-vim plugin."""
+class InvalidJavaPathError(OSError):
```
ChrisCoffey: 👍

ches:
Rebased, realizing this had minor conflicts. Would be nice to get this out of the way before refactoring to branch for ENSIME v2 gets underway.

ChrisCoffey:
LGTM

jvican:
LGTM.

PR #295, opened 2016-06-30
Fix broken request and missing import in debugger

ches:
The debugger module references the `json` module without importing it, causing an error to be thrown when that code path runs. Also, the `DebugClearAllBreaksReq` request was misspelled so it did not work since the server didn't understand it.


jvican:
As this is a big issue and a rather easy fix, i merge it right now.

PR #301, opened 2016-07-04
Code branching to support server 2.0. Closes #300

### LINKED ISSUE #300: Support server 2.0
Server 2.0 is starting to diverge from 1.0. It would be nice to support it somehow on the vim plugin side, maybe a separate branch (leaving 1.0 as default branch), so advanced users and contributors can easily switch (configuring the branch in their plugin manager) to start testing/developing the new features/API.

Keep support for both mayor versions in the same branch could lead to a lot of trouble and boilerplate.

ktonga:
@ches @ChrisCoffey @jvican Hey guys, it is just a WIP PR to have some early feedback from you as I'm not a Python dev and I have my OOP skills a bit rusty.

Is something like this what you had in mind for code branching?

If you ask me, I'd prefer to implement it just with curried and composed functions :-) but dunno if you wanna depend on PyMonad :-P


ensime_shared/symbol_format.py
```diff
@@ -1,20 +1,20 @@
 # Funcions for symbols formatting
 
-def completion_to_suggest(completion):
+def completion_to_suggest(completion, formatted_type_params):
```
jvican: It would be great to write a comment stating that `formatted_type_params` is a function.

jvican:
I think this is great. We all are functional programmers, but let's try to keep our functional pride and the amount of dependencies as small as possible :wink:. This is idiomatic python, so great! Now, we just need a more principled way to do this for the rest of the features. I would move the client to an independent file `client.py`, leaving only the class `Ensime` and then create a new file `client2.py` that holds all the new code for server `2.0`. This will help modularity and maintainability in the long run.

ensime_shared/symbol_format.py
ktonga: Will do

ktonga:
Sure! I was thinking the same.This file is getting huge, great we agree on that, I feel more confident about doing such a big change now.
I will do the file split, also the proper launcher changes and will push for another round of feedback.
Thanks for the comments.

ches:
> I would move the client to an independent file `client.py`, leaving only the class `Ensime` in `ensime.py` and then create a new file `client2.py` that holds all the new code for server `2.0`. This will help modularity and maintainability in the long run.

I think @ktonga has started with the simplest and least invasive change to start discussion, but what I was thinking before when I mentioned mixins on #300 is a base `EnsimeClient` class that still holds all code that can most likely be shared between versions, i.e. the message queue management, logging setup, etc.—likely almost everything except for specific handlers that end up differing. Move existing handlers into a mixin for v1 and start a new one for v2.

ensime_shared/ensime.py
```diff
@@ -584,9 +584,13 @@ def handle_completion_info_list(self, call_id, payload):
...
         completions = payload["completions"]
         self.log("handle_completion_info_list: in")
-        self.suggestions = [completion_to_suggest(c) for c in completions]
+        self.suggestions = [completion_to_suggest(c, self.format_type_params()) for c in completions]
         self.log("handle_completion_info_list: {}".format(self.suggestions))
 
+    def format_type_params(self):
+        """NOP as it's not available for Server 1.0"""
+        return lambda _: ""
+
```
ktonga: I was in doubt about implenting it as an abstract method. So you prefer to have a common class between versions and implement concrete classes for both versions, instead of make v2 override v1 behavior. Is that correct? It seems clearer to me.

jvican: Yes, thanks for pointing that out @ches.

ches: Actually this particular case of `format_type_params` might be a bad example, maybe it is best to duck-type it like you've done here so that code like `def completion_to_suggest(completion, formatted_type_params)` can be part of the common interface with the same signature. Or maybe it could be in kwargs to be more flexible to both? I'm not considering the specifics *too* closely just yet.

But overall, yeah, composition usually proves better than inheritance—roughly what I had in mind is a bit of both:

- Extract existing handler stuff from `EnsimeClient` to a mixin, called something like `ProtocolHandlerV1` maybe?
- Have `ProtocolHandlerV2` inherit from that, in Scala terms `trait ProtocolHandlerV2 extends ProtocolHandlerV1`. It can add stuff for new features and override for protocol changes as needed. I suppose there ought to be an abstract `ProtocolHandler` defining the interface, it's a bunch of OOP boilerplate but it's probably the proper thing to do even if Python can't statically check compliance 😄 
- Instantiate an `EnsimeClient` at startup dynamically mixing in the appropriate `ProtocolHandler` at runtime.

Then `EnsimeClient` could catch `NotImplementedError` in a common place, to log/inform the user that a new feature isn't supported in ENSIME v1.

This would introduce a good deal more OOP machinery, but on the other hand I think it would be a big step for better testability too. The worst thing deterring me from adding more tests currently is the tight coupling between `EnsimeClient`, a Vim instance (`self.vim = vim`), the launcher, and the handler logic which is the most important thing to cover with tests. If we can get to the point that those handler mixins can be tested in isolation, perhaps with just a minimal mock `vim` interface like @ChrisCoffey started in the test harness, that would be huge to me.

These are just my thoughts, some of the "architecture" stuff that's been bouncing around my head even before this server v2 discussion. Keen to hear everyone else's thoughts.

ktonga: @ches I agree mostly on everything. 

Strongly agree on improving testability, when I implemented completion formatting, the first thing I did was to refactor methods into just module functions. Having to create a full client just to test a bunch of functions that can be tested in isolation is a big stopper for adding new tests.
Not sure about the abstract class for the handlers, but I guess that having the interface defined in a single place is something good, so kinda agree on that too.
Don't agree on the runtime mixin bit, I was taking a look and it seems that you need to use that `__metaclass__` thing, which in my opinion complicates the code for not clear win. We can instead have fixed mixins for v1 and v2 and instantiate one of them based on the `server2` flag.

I can simplify even more the type params formatting, moving the __NOP__ inside `completion_to_suggest` and turn it on based on the server version flag, no need for overriding any method for this very case, so it is a bad example for testing the code branching.
I think I'll close this PR and open one for the mixin refactor + server 2.0 support, and a simple one just for the type params formatting.

To recap, I'll factor out just the handlers to a separate set of mixins, let's say `handler.py` with the abstract `ProtocolHandler` class and the concrete ones for V1 and V2, I think we can keep them all in the same file for now.

What the others think about it?

ches: It is a little bit magical, but not metaclass hackery exactly:

```python
>>> class ProtocolHandler(object):
...     pass
...
>>> class ProtocolHandlerV1(ProtocolHandler):
...     pass
...
>>> class ProtocolHandlerV2(ProtocolHandler):
...     pass
...
>>> class OtherMixins(object):
...     pass
...
>>> class EnsimeClient(OtherMixins, object):
...     pass
...
>>> server_v2 = True
>>> def during_startup():
...     if server_v2 == True:
...         Protocol = ProtocolHandlerV2
...     else:
...         Protocol = ProtocolHandlerV1
...     EnsimeClient.__bases__ = (Protocol,) + EnsimeClient.__bases__
...
>>> during_startup()
>>> client = EnsimeClient()
>>> client.__class__.__bases__
(<class '__main__.ProtocolHandlerV2'>, <class '__main__.OtherMixins'>, <class 'object'>)
```

Some Pythonistas balk at almost any form of metaprogramming, but this isn't all that uncommon. But yeah, I suppose there's little gain over something without the magic, like:

```python
>>> class EnsimeClientV1(ProtocolHandlerV1, EnsimeClient):
...     pass
...
>>> class EnsimeClientV2(ProtocolHandlerV2, EnsimeClient):
...     pass
...
>>> def factory():
...     if server_v2 == True:
...         return EnsimeClientV2()
...     else:
...         return EnsimeClientV1()
...
>>> client = factory()
```

ChrisCoffey: I much prefer the factory approach. It's much clearer to an uninitiated Pythonista like myself.

ktonga:
I've reverted my initial commit, rebased on master and implemented what we disused.

It seems I will need to introduce the first plugin setting for server v2 flag. Any advice?

ches:
This is looking good so far! I haven't run it yet, I'll try soon.

> It seems I will need to introduce the first plugin setting for server v2 flag. Any advice?

I would think it can just use plain VimL in `plugin/ensime.vim` and with the setup that's in place already it should work without special/duplicated handling for Neovim, e.g.

```vim
if !exists("g:ensime_server_v2")
  let g:ensime_server_v2 = 0
endif
```

ensime_shared/protocol_handler.py
```diff
@@ -0,0 +1,249 @@
+import webbrowser
```
ches: Minor style suggestion: since file names become modules in Python, `protocol.py` could be nicer here, and a little more general if anything else ends up logically belonging here. Then imports read a little better:

```python
from ensime_shared.protocol import ProtocolHandlerV1
```

ensime_shared/ensime.py
```diff
@@ -1097,12 +925,18 @@ def client_for(self, config_path, quiet=False, create_classpath=False,
         if abs_path in self.clients:
             client = self.clients[abs_path]
         elif create_client:
-            launcher = EnsimeLauncher(self.vim, config_path)
-            client = EnsimeClient(self.vim, launcher, config_path)
+            launcher = EnsimeLauncher(self.vim, config_path, self.server_v2)
+            client = self.do_create_client(launcher, config_path)
```
ches: It looks like you could move the launcher instantiation into `do_create_client()` now, simplifying what's passed around.

ensime_shared/protocol_handler.py
```diff
@@ -0,0 +1,249 @@
...
+        """Get a registered handler for a given response and execute it."""
+        self.log("handle_incoming_response: in {}".format(payload))
+        typehint = payload["typehint"]
+        handler = self.handlers.get(typehint)
+        def feature_not_supported(m):
+            msg = feedback["handler_not_implemented"]
+            self.raw_message(msg.format(typehint, self.launcher.ensime_version))
+        if handler:
+            with catch(NotImplementedError, feature_not_supported):
+                handler(call_id, payload)
```
jvican: I don't agree, actually. The `catch` sugar is way more clear than a simple try-catch expression, since it clearly states what is doing, without the syntactic overhead of writing try catch all the time.

ensime_shared/protocol_handler.py
```diff
@@ -0,0 +1,249 @@
...
+
+            if not url.startswith("http"):
+                url = gconfig["localhost"].format(port, payload["text"])
+
+            browse_enabled = self.call_options[call_id].get("browse")
+
+            if browse_enabled:
+                log_msg = "handle_string_response: browsing doc path {}"
+                self.log(log_msg.format(url))
+                try:
```
jvican: I would use `catch` here.

ensime_shared/protocol_handler.py
ches: &lt;slight rant&gt;

> without the syntactic overhead of writing try catch all the time

I would argue that writing throwaway functions, or Python's ugly lambdas, is not significantly lighter. It makes me think more about what error handling code is trying to do, not less.

> having the error handling functions independent of the code promotes reusability and better modularity.

I cited understanding of that intent when it's appropriate, and in this particular example, the handling code *is* possibly reusable, but as written so far it's a local function so it clearly isn't reused yet. I think much more than half of the `catch` usage in the code base uses a one-time handler. If this handler above were defined at a level where it could be reused, there's no reason it couldn't be called as the one-line body of an `except` clause for the cost of one or two extra newlines.

For comparison:

```python
def feature_not_supported(m):
    msg = feedback["handler_not_implemented"]
    self.raw_message(msg.format(typehint, self.launcher.ensime_version))

if handler:
    with catch(NotImplementedError, feature_not_supported):
        handler(call_id, payload)
else:
    self.log(feedback["unhandled_response"].format(payload))
```

versus

```python
if not handler:
    self.log(feedback["unhandled_response"].format(payload))
    return

try:
    handler(call_id, payload)
except NotImplementedError:
    msg = feedback["handler_not_implemented"]
    self.raw_message(msg.format(typehint, self.launcher.ensime_version))
```

Personally I find the latter much more immediately obvious. Same number of lines, FWIW. In fairness, the former could also use an early return.

The helper function is forced to accept an argument that it doesn't need/use, making me and linters wonder if there's a bug. Also, `catch` only passes the message string of the exception object to the callable, so it's limited in a case where there is something else useful about the object. If the handler were defined non-locally, I might need to navigate away to go verify what it does. In a case like this I think that cost outweighs two lines of handler code.

`try` is a very well-understood native language feature, so it has almost zero cognitive load. `with catch` is a clever custom thing of this project that one needs to learn, and as noted above it has limitations.

Ultimately I yield to @jvican's judgment, but several instances of `catch` having been irking so I had to state my case vocally this once…

ensime_shared/protocol_handler.py
ches: Heh, I originally wrote this, feeling a local function was worse 😇   Note that this handling code uses the exception object, which `catch` can't. For logging the full `__str__` of an exception is more helpful than only its message.

ensime_shared/protocol_handler.py
```diff
@@ -0,0 +1,249 @@
...
+                                            p["line"],
+                                            str(sym["name"]),
+                                            "info")
+                qfList.append(item)
+        self.write_quickfix_list(qfList)
+
+    def handle_symbol_info(self, call_id, payload):
+        """Handler for response `SymbolInfo`."""
+        warn = lambda e: self.message("unknown_symbol")
+        with catch(KeyError, warn):
```
ches: Sorry to beat a dead horse and this isn't @ktonga's code so we're veering into the weeds, but this is a `with catch` with a 20-line body. I'm scared to refactor it because I'm not very sure where exactly the `KeyError`(s) could come from.

To be fair you could just as easily write a `try` block this big, but I believe the code smell of that is more obvious in code review. So the (slightly) greater syntactic weight of `try`/`catch` sometimes has merit. Here it might give you more encouragement to stop and wonder, "could I eliminate this error handling entirely by using `dict.get()`?".

ensime_shared/protocol_handler.py
ches: Sorry, `catch` calls `str()` on the exception rather than getting it's message attribute, I was mistaken. I had to re-read the implementation.

ensime_shared/protocol_handler.py
ktonga: I don't have a strong opinion on this kind of small styling details, so at the end I will accept whatever you guys decide. But I think that changing it just in one place makes no difference, if we decide to change the error handling style it would be better to do it project-wide in a separate PR.

ensime_shared/protocol_handler.py
ktonga: This is just moved code, so I'd prefer to let it as it is. And in any case we decide to unify the error handling style do it in a dedicated PR

ktonga:
This isn't a WIP anymore, I think it's getting closer to be ready if you don't have any more comments on my last commits.
Also I think we should leave this setting undocumented for a while until we're sure it doesn't cause any trouble. Do you agree?

ches:
*[Side note: let's please not squash-merge this branch, when there are multiple authors it destroys record of that and ruins `git blame` for finding out who might be the best person to ask about a particular change].*

> Also I think we should leave this setting undocumented for a while until we're sure it doesn't cause any trouble. Do you agree?

Sounds good to me, especially because of the issue you noted in Gitter: an already-existing bootstrap project for the Scala version will be detected so the launcher won't know to replace it with a v2 install.

ensime_shared/protocol_handler.py
jvican: I disagree. I find very useful to split error handling logic into different local functions, even if they cannot be immediately reusable. IMO, I find it the right way to do it from a modularity standpoint (and following fp good practices). I understand this is not idiomatic of Python, but I see several benefits of wrapping everything into `catch`:

- It's damn easy to add more error handling statements without polluting what a function really does. Usually, when browsing code, we don't care about what happens if there's a failure, especially if we're learning a codebase. We just want to know how the logic proceeds. It's very annoying when I have to browse a codebase and I find 25-lines in the `catch` block that prevents me from focusing on the real important parts.
- The error handling logic can be reused in most of the times. This is not true for very concrete/idiomatic cases, for instance when catching `KeyError` or `NotImplemented`, but I consider healthier to go for other alternatives. Concretely, if a value is not in a dictionary, I prefer `get` instead of just expecting an exception to happen.
- One can use `catch` not to only execute handlers, but enrich the try-catch functionality with logging *without anyone noticing*. It is not only useful for debugging purposes, but also to check when some weird conditions happen and they shouldn't.
  
I don't know how people are using `catch`, but I would use it only to wrap statements that produce an exception. In fact, in case that more than one happens, it should wrap the minimum possible lines that cover all of them.

ches:
👍 

Finally got a chance to try running it today, sorry for delay. I have a few little ideas for enhancement but I think they can all wait for later:

-

ches:
(Keyboard misfire, sorry. Previous comment updated).

ches:
Actually since we already support [overriding with an assembly jar][1], we might not need any fancy futzing around in the plugin code to support more convenient development on different versions. Just the `g:ensime_server_v2` Boolean flag to switch which protocol handlers are loaded could be enough to carry us a long way.

Anyway, this kind of discussion can come on a new follow-up issue.

[1]: http://ensime.github.io/contributing/#manual-qa-testing

ches:
@jvican @ChrisCoffey Any concerns here? I've tested this but I haven't run it consistently because I've been working on other branches. This PR has a big surface area so I'd hate for @ktonga to get stuck rebasing repeatedly or for other people to do work that has messy conflicts after this lands.

@ktonga have you been running this in daily work for awhile?

I know some features are broken when using server v2, that's expected, fixing them is out of scope for this PR.

jvican:
LGTM!

ktonga:
That was a relief. This PR was starting to burn.

@ches I have been using it while hacking on ensime/ensime-server#1471 and I've just opened a ticket (#313)  for the only problem I found, which is only visible when you turn `v2` on.

jvican:
Next time let's wait for @ChrisCoffey's feedback, I know sometimes we're slow in reviewing but it's good if we all are aware of the new changes in the repo :smile:.

ktonga:
Ok, will do, sorry. I thought that two LGTMs were enough.

jvican:
Yes, usually. But we haven't talked about merge policy yet and @ches got push rights recently. I guess it's time for you to decide :+1:

PR #302, opened 2016-07-06
Linter and autoformatting config and fixes

ches:
We've had flake8 and autopep8 in the dev dependencies, but their advice hasn't really been getting followed 😊   That may be partly because the `make` target recipes weren't quite right, `**/*.py` probably only works on Zsh.

I haven't actually applied any fixes on this branch yet, because I figured there may be some discussion about settings and they would just make a big noisy diff in the meantime. Also if anybody has any big WIP (@ktonga) we could wait on a mass formatting PR until it won't create annoying conflicts.

If you're not familiar, briefly:

flake8 is a wrapper tool for a few other linter-type tools, that also adds plugin extensibility (including the `pep8` tool, which is in the process of being renamed to `pycodestyle`…). flake8 is really widely used in quality Python projects so IMO using it is a no-brainer.

flake8 just reports issues, `autopep8` actually fixes them (the PEP 8 style ones, anyway). I'm always somewhat leery of auto-formatters but it seems to mostly do a good job, I ran it locally with the setup on this branch and about the only ugly thing it did that a human should probably improve is this:

```diff
     def handle_import_suggestions(self, call_id, payload):
-        imports = list(sorted(set(suggestion['name'].replace('$', '.') for suggestions in payload['symLists'] for suggestion in suggestions)))
+        imports = list(sorted(set(suggestion['name'].replace('$', '.') for suggestions in payload[
+                       'symLists'] for suggestion in suggestions)))
```

The rest of the changes were pretty much what you'd expect. Let me know what you all think about using it. If it leaves 2% for a human to fix it's probably still useful. I can push the automatic fixes to this branch to see if you like.

Right now options are mostly the tool defaults, we can tweak them over time to turn off some errors if we don't like them. Run `make lint` yourselves and see what you think. (Then try `make format` and then lint again).

After some time goes by and we settle on optimal settings for the maintainers' tastes, we could push adherence to the standards more aggressively, like possibly making style warnings part of CI for automatic pull request checks.

By the way, if you `pip install flake8` with the default Python on your path, Syntastic should support it automatically (or probably starting Vim from the shell where you've activated this project's virtualenv would work too).


setup.cfg
```diff
@@ -0,0 +1,36 @@
+# Note: setup.cfg is normally a config file for Python package publishing. In
+# our case it's just a shared config file for linting and formatting tools
+# because most of them support it automatically.
+
+[flake8]
+# Try to stick to 79, but sometimes being religious *hurts* readability.
+max-line-length = 100
```
ches: About the only non-default so far—I am a firm believer in keeping line lengths under control but often feel like the PEP 8 standard of 79 ends up leading to a lot more annoyance and awkward readability than when it's raised just a bit.

Open to your input.

ches:
I thought that most of [OpenStack's Python coding standards][1] are pretty good. They have a flake8 plugin that adds checks for many of them.

[1]: http://docs.openstack.org/developer/hacking/

ktonga:
Big yes to this. The last time I touched the `Makefile` it surprised me that we had those tools "ready" to use but they were not used in the CI setup, so I was close to open a PR to enable them, much better that you bring it up as you are miles ahead with you Python knowledge and all the things you propose sound simply great.
And thank you very much for not raising it right now, it could be a hell of a merge for my WIP PR.

setup.cfg
ChrisCoffey: 100 is about the max I'd want to use, but it's fine w/ me.

test-requirements.txt
```diff
@@ -1,9 +1,11 @@
...
-lettuce==0.2.21
+# === Test ===
+lettuce~=0.2.21
 
 # Optional speed-up for Lettuce, but Docker image build issue needs fixing, see:
 # https://github.com/ensime/ensime-docker/pull/8
 # python-Levenshtein~=0.12
+
+# === Dev Tooling ===
+flake8
```
ChrisCoffey: is there a reason not to pin the version?

setup.cfg
ches: Could try a bit lower. It'd be nice if the tools had a notion of "soft limit" and "hard limit" with warning vs. error.

test-requirements.txt
ches: I just thought releases of these are probably unlikely to break things and it's not for production code, so it's nice to get updates occasionally without thinking about it. But I can restore them if you think it's best, if used in CI some constraint is a good idea. The `~=` "compatible" operator on the minor version could be a good compromise (basically expect Semantic Versioning).

setup.cfg
jvican: I'd go with the standard 80.

jvican:
What's missing to merge this?

ches:
> What's missing to merge this?

I think just your LGTM 😄 

And probably consensus on two fairly minor points:

- Whether we care about pinning the tool dependency versions. Makes sense for CI when we get to that. Strict or Semantic?
- Max line length. I completely agree that 79 is *usually* best to strive for but sometimes it becomes onerous and makes readability suffer if the tools strictly enforce it. I wish the they supported warning/error soft and hard limits, but alas no. 100 gets used by a lot of projects with good code standards, e.g.:
  - [Requests](http://docs.python-requests.org/en/latest/dev/contributing/#kenneth-reitz-s-code-style)
  - [Coverage.py](https://bitbucket.org/ned/coveragepy/src/00572535a07db5ddcc3ede80101921f1f1db646b/tox.ini?fileviewer=file-view-default#tox.ini-93)

jvican:
I'm very strict with the max line length, I usually use 80. But I understand that other people are more lax on this, so I'll let you decide whichever you think it's better.
  
Regarding to the versioning, I don't personally mind. I think that pinning the versions would be pretty safe in case they update and break source compatibility. What do you think?
  
Let's quickly discuss about it and merge it. We need to dequeue some of the PRs :wink:.

ches:
Okay, I've pinned the versions with `~=`, I think that's a good compromise all around. Only talking about test/tooling deps here anyhow, I'd agree with strict for production code deps. If any of the tools ever break our CI in a minor release, we will take away their trusted privilege swiftly 😁 

About line length, a possible compromise to approximate warning/error limits: leave the linter failure setting at 100, but set Vim modelines with `set colorcolumn=79` to encourage the soft limit. This is a Vim plugin project so it's a reasonable assumption contributors are using Vim and don't find modelines unwelcome. A few people do disable them completely, but there's still code review and hopefully a hard limit in CI…

Just an idea.

jvican:
It sounds good, let's go for it!

ches:
Amended one more time just adding `virtualenv` to the dev requirements, because it is required by the makefile. There's an equivalent tool built-in to Python 3 installs but we're not there yet…

PR #308, opened 2016-07-14
Bail out with a warning message when no Python available

### LINKED ISSUE #307: Fail with a useful message when Python support isn't available in Vim
We ask users to confirm that their Vim is built with `+python` prominently at the start of our installation instructions on the web docs, but if they don't, they end up getting ugly errors when trying to run `:EnInstall`. We can easily give a nicer error message with a `has('python')` check and bail out early. I believe Neovim works exactly the same way if the Python remote plugin host hasn't been installed, I'll try.

References #306

ches:
Tested this on Neovim after doing `pip uninstall neovim`. I don't have a standard Vim build at hand built *without* `+python` but I think it's pretty safe to say this should work.

Closes #307

### DIFF
--- a/plugin/ensime.vim
+++ b/plugin/ensime.vim
@@ -1,6 +1,20 @@
-if has('nvim') | finish | endif
 if exists('g:loaded_ensime') || &cp
     finish
+else
+    if !has('python')
+        echohl WarningMsg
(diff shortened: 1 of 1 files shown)


jvican:
LGTM

Issue #136, opened 2015-12-22
Fix unit tests


yazgoo:

jvican:
OK, I think it could be better if we only make tests for the neovim plugin (in python). It's really easy to embed a neovim instance and check functionality against a real instance of the editor. Vim won't be covered by the tests, but as all the logic of the plugin is in ensime_shared (which is shared by both vim and neovim) vim will indirectly be covered by the tests.
  
What do you think? More info [here](https://github.com/neovim/python-client).

jvican:
Any feedback on this?

ChrisCoffey:
How do you feel about using a Cucumber dialect like [Lettuce](http://lettuce.it/tutorial/simple.html#tutorial-simple) so we can share tests between emacs & vim? 

I'm pretty excited by the prospect and would love for our CI to work again, so I'm willing to take this on.

jvican:
I think it's a great idea that I'd like to see implemented. I actually looked into this when Sam commented it, but I didn't see any feasible way of integrating it into `ensime-vim` in a clean way. If you can find it, please go ahead, this would be probably one of the best features of ensime-vim by far.

jvican:
Hey @ChrisCoffey, just checking in, are you doing progress on the new test suite?

ChrisCoffey:
I've been poking around with it, but was on vacation last week, so no real progress. At this point my plan is to chop this up into smaller tasks so folks can start picking them off.

jvican:
That’s a great idea. Keep up the good work :smile:

ches:
Okay, we've had the Lettuce suite set up for awhile, #327 added unit-level testing, and #312 proposes an implementation of embedded Neovim integration tests as @jvican suggested above. I think we can safely close this now 😄

PR #312, opened 2016-07-17
Integration testing setup with server install / launcher tests

ches:
*(Note: this will certainly fail the build until I can update the Docker image with a Neovim installation)*

This has a few rough edges which I'll point out inline, but it's ready to poke around with and discuss.

I have added a test setup that integrates against a real Neovim instance, with initial tests of the launcher functionality as a proof of concept. @jvican mentioned this before (https://github.com/ensime/ensime-vim/issues/136#issuecomment-175908679) and I think it's a great idea—Gherkin BDD is really optimized for "full-stack" tests (acceptance, end-to-end, whatever). This isn't 100% "black box" since using the Neovim Python client is RPC versus actually driving a UI, but it's close enough for the depth of coverage we get with such an easy setup, and it can run completely headless for CI. Of course this won't cover the VimL code that's specific to standard Vim—as @jvican noted—but that's a thin layer, and as far as we can trust Neovim's compatibility it comes pretty close since the tests exercise things by executing commands (e.g. `:EnInstall`), etc.

Things like vimrunner that can script Vim execution with `+client-server` seem unlikely to be as robust, and I think it's safe to say we would never want to maintain a CI setup that has X and a virtual framebuffer config, so this is probably the best we're reasonably going to do.

One caveat to try to resolve: running the tests sometimes leaves ENSIME server processes running, especially in the headless/embedded mode. I'm not sure why yet.

Probably obvious, but you need Neovim installed with `nvim` on your path, and sbt.

I made an asciicast, because hey, it's kind of fun to see 😄   Running on a fairly wimpy MacBook Air so the sbt runs are not the fastest.

[asciicast](https://asciinema.org/a/80015)


ensime_shared/spec/features/ensime_harness.py
```diff
@@ -1,23 +0,0 @@
...
-
-class TestVim:
-    def __init__(self):
-        self.current = TestVimCurrent()
-
-    def command(self, what):
-        return None
-
-    def eval(self, what):
-        return "/tmp"
```
ches: These were unused at this point, and I think they're most likely to belong in unit tests instead of integration. Stay tuned for some thoughts/work I have toward that.

ensime_shared/spec/features/server_bootstrap.feature
```diff
@@ -0,0 +1,18 @@
...
+    When I edit a Scala file
+    And invoke server installation
+    Then ENSIME should be installed
+    And the server should be started
+
+  Scenario: Begin an editing session with server installed
+    Given I have installed ENSIME server
+    And I have created a valid .ensime project config
+    When I edit a Scala file
+    Then the server should be started
```
ches: cc @fommil

ensime_shared/spec/features/server_bootstrap_steps.py
```diff
@@ -0,0 +1,66 @@
...
+from lettuce import step, world
+
+from ensime_shared.launcher import EnsimeLauncher
+
+
+@step('no server bootstrap project exists')
+def recreate_boostrap_project_path(step):
+    bootstraps = world.bootstrap_projects_path
+    shutil.rmtree(bootstraps)
+    os.mkdir(bootstraps)
```
ches: This sandbox for bootstrap projects doesn't actually work yet, the suite as of now will use your `~/.config/ensime-vim`. Ideally it wouldn't, but this works for now.

`EnsimeLauncher` is parameterized already to allow a different path for the bootstrap project to be used, but it's hard to utilize that unless we were unit testing—can't easily pass a specific constructor param to a class at this level of testing. I think the sensible way to support this would be to allow setting the path with a `g:` variable, which we could set for tests. This could actually be useful for some users and other development/testing needs as well anyway. I could work on adding support for this on this branch or later.

ensime_shared/spec/features/terrain.py
```diff
@@ -0,0 +1,115 @@
...
+def reset_vim(_scenario):
+    # Boilerplate to set up embedded Neovim test instance, by default, allowing
+    # separate process optionally.
+    if LISTEN_ADDRESS:
+        world.vim = neovim.attach('socket', path=LISTEN_ADDRESS)
+    else:
+        embed_cmd = ['nvim', '-u', world.vimrc, '-i', 'NONE', '--embed']
+        world.vim = neovim.attach('child', argv=embed_cmd)
+
+    world.vim.input(CLEANUP_FUNC.format(vimrc=world.vimrc))
```
ches: This is mostly where rough edges surface.

The [neovim/python-client test suite][2] which I pillaged for ideas sets up the `vim` instance *once* at the beginning of the suite, and calls the state reset function per-test. I tried this, and it worked fine in non-embedded mode, but with embedding it just hung consistently, very early in the test run. I didn't figure out why yet. Changing it to `@before.each_scenario` worked around that. Lettuce's `world` is shared global state, so it might be problematic for concurrent code. It would of course be more performant to do it once, though with only a few tests so far I believe it's negligible.

The Python client's tests also don't need to test any plugins like ours do, so there are some differences. AFAIK there is currently no way to *reload* a remote plugin in a Neovim instance, like you might be able to do with a VimL plugin by un-`let`ing include guard vars, removing autocommands, and doing `runtime plugin/foo.vim` afterward. I asked in Neovim Gitter about this. Making a new `vim` instance per-scenario as done here probably obviates the need for that anyway, but I'm not entirely confident of my understanding of the whole remote plugin host architecture yet, so I'm not certain.

I also have a hunch that changing out the `vim` instance like this might be a source of the ENSIME server processes getting left behind after test runs—perhaps Python garbage collection steps in without the `VimLeave` getting run.

I've left the `BeforeEachTest` function infrastructure in place here in case we can figure out a way to move the `vim` instance setup back into `@before.all`, maybe others want to take a stab at it when trying this branch. It is probably unnecessary as it stands now, with the new `vim` instance per-test.

Also worth mentioning, there is [a `remote#host#LoadRemotePlugins` function][1] in the Neovim API. When I had the `vim` instantiation in `@before.all`, I tried calling that in the `BeforeEachTest` function at the end after re-`sourcing` the vimrc, and that worked to accomplish sanitizing the session state and then loading our plugin, but *only when running a single scenario* with `lettuce -s [N]`. Other code in that Neovim API throws an error when a remote plugin gets "registered" more than once when executing multiple tests 😞  This is an area where the Neovim team is probably still open to ideas, but of course it's not likely to change overnight.

[1]: https://github.com/neovim/neovim/blob/7fa1baf44e7877b7b6b861c7140b006d4cee79f3/runtime/autoload/remote/host.vim#L130
[2]: https://github.com/neovim/python-client/blob/master/test/test_common.py

ches: > I also have a hunch that changing out the `vim` instance like this might be a source of the ENSIME server processes getting left behind after test runs—perhaps Python garbage collection steps in without the `VimLeave` getting run.

Yeah, if I do something like this it seems to take care of the orphan processes problem:

```python
# If we're re-creating the vim instance, make sure to properly fire
# VimLeave so that ENSIME process is shut down
if hasattr(world, 'vim'):
    world.vim.quit()
```

But it introduces failure that's probably a race condition. I'll see what I can work out.

One idea is, like the setting variable for where bootstrap projects live, that this might suggest new user interface could be desirable. If we had commands to stop/start/restart ENSIME server, the test suite could use them and get better determinism. Users might occasionally want them if something seems to be wonky.

jvican:
Hey @ches, the work on this is awesome! What's the status of this PR? Is there something that could be done by an ensime-vim contributor? Do we only need to update the Docker image? It seems that @jqno could do some progress on adding tests for the autocompletion feature.

fommil:
Is it hard to add neovim to the base ensime image? ensime-docker v2.x

ches:
> Is it hard to add neovim to the base ensime image?

It shouldn't be, I just never got around to it in part since I don't think anyone else tried this out until now.

> ensime-docker v2.x

…but that may (or may not?) be a rub. [Last I remember][1] there was some sticky situation about rebuilding the branch of the Docker image that has the server pinned at v1.x. I don't think I ever got a complete understanding of the scheme of how the images are built/tagged from that thread (e.g. manual build-and-push vs. automated builds IIRC), which is also probably another reason I didn't mess with the Docker image here yet…

[1]: https://github.com/ensime/ensime-docker/pull/8

fommil:
I think the tldr is: send it to the v2.x branch of the docker image.

ches:
> What's the status of this PR? Is there something that could be done by an ensime-vim contributor?

@jvican It probably needs a rebase, but it should be a pretty easy one. There was still some fragility in this too (some notes in my hidden comments above, if they make any sense).

Mainly it's just that we never got into any discussion about this PR. I whipped up the branch on a whim when I wanted to play with the Neovim embedded possibility and had some time. I still think an integration testing harness could be useful in the long run, though this initial case (launcher) might be less interesting for the future after #354. If we can get it to be stable though it still serves as an example of the framework that others can follow for other functionality (completion sounds like it could be a good one).

ches:
> I think the tldr is: send it to the v2.x branch of the docker image.

K, I'll give it a try when I can give it a few minutes' attention. Our builds are currently running against `ensime/ensime:v1.x-cache` but there might be nothing that breaks on v2.x anyway…

fommil:
I strongly recommend upgrading to the v2.x-cache images. If you want to use java 6, make sure to have a `.java-version` file containing `1.6`

Issue #309, opened 2016-07-14
Clean lint bill of health, add to CI


ches:
This will wait on #301 to avoid merge conflict pain, but otherwise since #302 is landed I'll create a PR to get `make lint` running clean so we can consider adding it as a CI check.

Also, on #302 I proposed this potential compromise for soft/hard line limit warnings/errors:

> …a possible compromise to approximate warning/error limits: leave the linter failure setting at 100, but set Vim modelines with `set colorcolumn=79` to encourage the soft limit. This is a Vim plugin project so it's a reasonable assumption contributors are using Vim and don't find modelines unwelcome. A few people do disable them completely, but there's still code review and hopefully a hard limit in CI…
>
> Just an idea.

If others like that idea then I can include it in the PR, alongside the boilerplate `# coding: utf-8` comments that will go in with #301.

ches:
Closed via 2644ec3e3a5bf8558e

PR #314, opened 2016-07-18
Clean lint bill of health, format source

ches:
As per #309, this runs `make format` with a little bit of manual help on some of the gnarliest code, and takes minimal steps to get `make lint` to exit 0.

I've also tried adding to the CI setup, we can discuss that further. ~~Not certain if my initial configuration is going to work anyway~~, I haven't tried [local Drone testing][1] yet—to my understanding the Drone CLI has changes in the latest version, I'm not sure what version our CI infrastructure runs but I believe it isn't the latest yet so I don't know what CLI to install… **Edit:** Seems to work 😄 

See also the "soft limit" line length idea on #309 and let me know what you think of that, haven't done it yet.

[1]: http://readme.drone.io/devs/cli/#local-testing:b659b046131d4024ab5e2d3675716bf0

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,7 +1,21 @@
 build:
-  image: ensime/ensime:v1.x-cache
-  pull: true
-  commands:
-    - pip install virtualenv python-coveralls
-    - make test
+  quality:
+    image: ensime/ensime:v1.x-cache
+    pull: true
+    commands:
+      - pip install --upgrade virtualenv
+      - make lint
+      - make format ;
+        if $(! git diff --exit-code --quiet) ; then
+          echo "Code formatting does not meet the project's standards:" ;
+          git --no-pager diff ;
+          exit 1 ;
+        fi
 
+  test:
+    image: ensime/ensime:v1.x-cache
+    commands:
+      - pip install --upgrade virtualenv
+      - make test
+    when:
+      success: true
--- a/ensime_shared/config.py
+++ b/ensime_shared/config.py
@@ -16,7 +16,8 @@ feedback = {
     "failed_refactoring": "The refactoring could not be applied (more info at logs)",
     "full_types_enabled_off": "Qualified type display disabled",
     "full_types_enabled_on": "Qualified type display enabled",
-    "handler_not_implemented": "The feature {} is not supported by the current Ensime server version {}",
+    "handler_not_implemented":
+        "The feature {} is not supported by the current Ensime server version {}",
     "indexer_ready": "Indexer is ready",
     "invalid_java": "Java not found or not executable, verify :java-home in your .ensime config",
     "manual_doc": "Go to {}",
@@ -43,7 +44,7 @@ commands = {
     "current_file": "expand('%:p')",
     "until_last_char_word": "normal e",
     "until_first_char_word": "normal b",
-    # Avoid to trigger requests to server when writing
+    # Avoid triggering requests to server when writing
     "write_file": "noautocmd w",
     "input_save": "call inputsave()",
     "input_restore": "call inputrestore()",
@@ -55,7 +56,8 @@ commands = {
     "split_window": "split {}",
     "vert_split_window": "vsplit {}",
     "new_vertical_window": "{}vnew {}",
-    "new_vertical_scratch": "{}vnew {} | setlocal nobuflisted buftype=nofile bufhidden=wipe noswapfile",
+    "new_vertical_scratch":
+        "{}vnew {} | setlocal nobuflisted buftype=nofile bufhidden=wipe noswapfile",
     "doautocmd_bufleave": "doautocmd BufLeave",
     "doautocmd_bufreadenter": "doautocmd BufReadPre,BufRead,BufEnter",
     "filetype": "&filetype",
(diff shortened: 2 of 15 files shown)


ensime_shared/ensime.py
```diff
@@ -49,7 +47,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
     which stores the a handler per response type.
     """
 
-    def __init__(self, vim, launcher, config_path):
+    def __init__(self, vim, launcher, config_path):  # noqa: C901 FIXME
```
ches: Tools report that the cyclomatic complexity of `EnsimeClient.__init__` is too high—I completely agree, but refactoring will be a significant job for another day…

ensime_shared/ensime.py
```diff
@@ -156,21 +155,23 @@ def queue_poll(self, sleep_t=0.5):
...
             if self.ws:
-                def logger_and_close(m):
-                    self.log("Websocket exception: {}".format(m))
+                def logger_and_close(msg):
+                    self.log("Websocket exception: {}".format(msg))
                     if not self.running:
                         # Tear down has been invoked
                         # Prepare to exit the program
-                        connection_alive = False
+                        connection_alive = False  # noqa: F841
```
ches: Linter understandably thinks this is an unused local variable—this *is* pretty tricky code, in part this is a symptom of `with catch()` encouraging this deeply nested local function…

ches:
GitHub tricks reminder: [diff omitting whitespace-only changes](https://github.com/ensime/ensime-vim/pull/314/files?w=1).

ensime_shared/config.py
```diff
@@ -55,7 +56,8 @@
     "split_window": "split {}",
     "vert_split_window": "vsplit {}",
     "new_vertical_window": "{}vnew {}",
-    "new_vertical_scratch": "{}vnew {} | setlocal nobuflisted buftype=nofile bufhidden=wipe noswapfile",
+    "new_vertical_scratch":
+        "{}vnew {} | setlocal nobuflisted buftype=nofile bufhidden=wipe noswapfile",
```
ches: This is somewhat out of scope, but since it relates to style it's probably appropriate to bring it up now: I don't know how I feel yet about the `commands` dictionary in general, for `feedback` I think it's sensible and the fact that it's been used thus far would make it pretty easy to add localization to our plugin, a beneficial side effect.

For commands, I wonder if it would be better to encapsulate them as methods on some kind of `Editor` object instead. That would probably help a good deal with testability, be more readable than `self.vim.command(commands['foo_bar'].format("whatever"))` sprinkled all over the code base, and would provide nice go-to-def and autocompletion with argument hints via jedi-vim 😄 

Whether or not we try that, or in the interim, I strongly wish for all of these format strings to require keyword parameters, e.g.

```python
commands = {
    # ...
    "split_window": "split {filename}",
    # ...
}

# call site
self.vim.command(commands['split_window'].format(filename="foo.scala"))
```

Short of actual object methods, that at least makes things a bit less opaque as you're reading code IMO, and less error-prone.

ensime_shared/ensime.py
```diff
@@ -674,7 +676,9 @@ def send_refactor_request(self, ref_type, ref_params, ref_options):
 
     def apply_refactor(self, call_id, payload):
         """Apply a refactor depending on its type."""
-        if payload["refactorType"]["typehint"] in ["Rename", "InlineLocal", "AddImport", "OrganizeImports"]:
+        supported_refactorings = ["Rename", "InlineLocal", "AddImport", "OrganizeImports"]
```
ktonga: Good guess-best-intermediate-variable-name logic for a formatting tool :stuck_out_tongue_winking_eye:

.drone.yml
```diff
@@ -1,7 +1,21 @@
...
-    - make test
+  quality:
+    image: ensime/ensime:v1.x-cache
+    pull: true
+    commands:
+      - pip install --upgrade virtualenv
+      - make lint
+      - make format ;
+        if $(! git diff --exit-code --quiet) ; then
+          echo "Code formatting does not meet the project's standards:" ;
```
ktonga: This might be clarified somewhere in the Contributing Guidelines to avoid frustrating rejections.
Could it be enforced in a Git hook maybe? Sounds like a job for machines :)

ches: The trouble with Git hooks is we have to ask everyone to install them locally. Seems unlikely to happen in a public project and the extra setup work might be *more* of a barrier to contribution than a CI failure.

I used this initially as what seemed to me "simplest thing that would work"—we could look at other things that might be a bit nicer experience for contributors, like [Scrutinizer][1] or [Landscape][2], but that's more stuff to set up and they use different tools & configuration than what is already in place so there would be a few changes (I'm not opposed to considering that route though).

I took this `echo` straight from ensime-server's main CI run by the way, although in that case Scalariform runs automatically on `sbt compile` so there's a good chance you would have noticed it before preparing a pull request. Agree that our expectations (and available `make` targets, etc.) should definitely be documented in contributing docs, however we end up implementing the checks.

[1]: https://scrutinizer-ci.com/docs/guides/python/automated-code-reviews
[2]: https://landscape.io/

ches:
Rebased to resolve conflicts. Any further thoughts about this? What do we need to move forward?

cc @ChrisCoffey

ensime_shared/config.py
ChrisCoffey: I like the idea of an editor object that exposes most of the vim functionality we use.

ChrisCoffey:
The drone checks were the only thing that would have prevented me from saying this is good to go. I'm comfortable landing this & seeing how it is received over the next couple weeks.

ches:
> The drone checks were the only thing that would have prevented me from saying this is good to go. I'm comfortable landing this & seeing how it is received over the next couple weeks.

Just curious, how do you think we could improve it?

I'll go ahead with this as-is then, will be good to have out of the way for other stuff in the works. We can of course tweak over time if anything about the build setup or specific linter settings proves onerous.

PR #316, opened 2016-07-19
Fix setting-exists check. Fixes #315

### LINKED ISSUE #315: E121: undefined variable: g:ensime_server_v2
It seems that the following line is not doing what I expected.

```VimL
key_exists = self.vim.eval("exists('{}')".format("g:ensime_server_v2"))
```

It's always returning `True` due to this funny behaviour in `vim` python module:

> vim.eval(str)						*python-eval*
>	...   Returns the expression result as:
>	- a string if the Vim expression evaluates to a string or number

ktonga:
The check was always returning `True` 'cause `vim.eval()` returns a string if the expression evals to a number and `bool('0')` is True

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -875,7 +875,7 @@ class Ensime(object):
 
     def get_setting(self, key, default):
         gkey = "g:ensime_{}".format(key)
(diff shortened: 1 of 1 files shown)


ches:
Doh, not sure how I didn't notice when testing the v2 branch.

LGTM, will merge since it's a small hotfix.

PR #317, opened 2016-07-19
Fix read `server_v2` setting. Same problem as in #315

ktonga:
Reading the flag value was affected by the same mistake as in checking its existence. If it was present, it was always `True`.

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -871,12 +871,18 @@ class Ensime(object):
         self.init_integrations()
 
     def init_settings(self):
+        """Loads all the settings from the ``g:ensime_*`` namespace.
+        
+        Invoked on client creation to avoid ``autocmd`` deadlocks.
+        """
         self.server_v2 = bool(self.get_setting('server_v2', 0))
 
     def get_setting(self, key, default):
-        gkey = "g:ensime_{}".format(key)
-        key_exists = int(self.vim.eval("exists('{}')".format(gkey))) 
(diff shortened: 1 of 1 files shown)


ches:
I just noticed like 10 minutes ago while working on something else that we might be able to avoid this issue more cleanly with the `vim.vars` property, here's a diff from my work tree:

```patch
@@ -877,10 +877,12 @@ class Ensime(object):
     def init_settings(self):
         self.server_v2 = bool(self.get_setting('server_v2', 0))

-    def get_setting(self, key, default):
-        gkey = "g:ensime_{}".format(key)
-        key_exists = int(self.vim.eval("exists('{}')".format(gkey)))
-        return self.vim.eval(gkey) if key_exists else default
+    def get_setting(self, key, default=None):
+        """Returns the value of a Vim variable ``g:ensime_{key}``.
+
+        If the variable is unset, returns ``default``.
+        """
+        return self.vim.vars.get('ensime_{}'.format(key), default)

     def init_integrations(self):
         syntastic_runtime = os.path.abspath(
```

Maybe give that a try, in a quick test when a var was set to an integer value, I actually got a Python `int` back.

ktonga:
Nice. Will give it a try. I might add some docs as well. Thanks.

ktonga:
@ches reworked based on your recommendation. I've tested all the possible cases and it seems to be working nicely.

ches:
LGTM 😊

ChrisCoffey:
LGTM as well. Thanks for adding the comments

Issue #294, opened 2016-06-29
"Disable plugin" mechanisms need to work completely and reliably


ches:
The `disable_plugin()` function is currently tightly coupled to printing messages about WebSockets errors, which probably ought to be changed so that it can be reusable in other situations where we can't recover from error conditions to initialize the plugin. It also just doesn't seem to really work for me—it needs to do additional things, like unloading autocommands, critically. It would probably be useful to look at [scriptease.vim] for some inspiration.

[scriptease.vim]: https://github.com/tpope/vim-scriptease

jvican:
Great idea, I'd love to see this working reliably.

PR #318, opened 2016-07-20
Improve portability with standard library functions

ches:
This is a pretty small change, some sundry things I noticed while trolling through the source. It got me started working on #294 since that's what `fetch_runtime_paths` is used for, so I was planning to package them up in a pull request together, but that's actually turning out to be a can of worms that touches the plugin core so I'm going to want to keep that review as focused as possible.

I don't know how close we are to running on Windows, but these were a couple of things that were easily improved by using the standard library. As the commit message says:

The `tempfile` module is Windows-friendly and will give us a secure temp directory, with permissions restricted to the user.

`os.path.expanduser` again supports Windows alternate idioms of `$HOME`.

Also, we should try to clean up our temp dir in teardown.

### DIFF
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -4,7 +4,9 @@ import datetime
 import inspect
 import json
 import os
+import shutil
 import sys
+import tempfile
 import time
 from subprocess import PIPE, Popen
 from threading import Thread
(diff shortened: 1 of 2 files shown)


ches:
This is a pretty small one in comparison to others I have mounting up… any remarks, @ChrisCoffey @ktonga?

ChrisCoffey:
nope, this looks great. I'm happy to see us actually creating proper temp files & cleaning up after ourselves. Feel free to land this.

PR #322, opened 2016-07-27
Curate completions list to avoid server bug. Fixes #319

### LINKED ISSUE #319: Sometimes completions come without `typeInfo` field
It happens some times, the case I was able to reproduce it for was completing class name/constructor:

```throw new IllegalSta|``` (`|` is the cursor position when I triggered the completion)

It shows the following error:

    Error detected while processing function EnCompleteFunc:
    line    1:
    error caught in request handler '/Users/tonga/Projects/ensime-vim/rplugin/python/ensime.py:function:EnCompleteFunc ([0, 'Illegal'],)':
      File "/Users/tonga/Projects/ensime-vim/rplugin/python/ensime.py", line 171, in fun_en_complete_func
        return super(NeovimEnsime, self).fun_en_complete_func(*args, **kwargs)
      File "/Users/tonga/Projects/ensime-vim/ensime_shared/ensime.py", line 859, in wrapper2
        return f(self, client, *args, **kwargs)
      File "/Users/tonga/Projects/ensime-vim/ensime_shared/ensime.py", line 1108, in fun_en_complete_func
        return client.complete_func(findstart, base)
      File "/Users/tonga/Projects/ensime-vim/ensime_shared/ensime.py", line 829, in complete_func
        self.unqueue(timeout=self.completion_timeout, should_wait=True)
      File "/Users/tonga/Projects/ensime-vim/ensime_shared/ensime.py", line 747, in unqueue
        self.handle_incoming_response(call_id, _json["payload"])
    Error detected while processing function EnCompleteFunc:
    line    1:
    KeyError: 'typeInfo'
    Pattern not found

For a completion request like this, the server returns some elements lacking the `typeInfo` field, the one we use to format the entries in the completion popup, so it fails.

Sample completion element:

```json
{
  "name": "IllegalStateException",
  "typeSig": {
    "sections": [],
    "result": "java.lang.IllegalStateException",
    "hasImplicit": false
  },
  "isCallable": false,
  "relevance": 40
}
```

We could easily avoid this error by skipping such elements, but it looks like a server bug. What do you think @fommil ?

CCing @Jacoby6000 who also experienced the bug

ktonga:


ches:
Is there an issue against the server we can track for this?

ches:
Otherwise LGTM.

ktonga:
As far as I can tell there is no issue, but I can search or create it and then link it in an ensime-vim issue to keep track.

ktonga:
@ches I've created both issues (server and vim plugin) to track the removal of this hack. Also updated the PR to reference the issue in the comment.

Hey @ChrisCoffey do think I can merge it? it's a small changes and fixes an annoying bug which is showing up quite often.

ches:
I'm running this without issue, I think it's pretty clear that it's safe. Let's get it in.

Issue #297, opened 2016-06-30
Overhaul logging and only do tracing when a debug flag is active


ches:
I think it's time to create a issue to track this as a TODO. The plugin's logging approach is pretty simplistic and heavy-handed—it's a homegrown `log` function that appends to a non-configurable file, re-opening a file handle on every call, and our logging is mostly trace-level detail that's quite verbose.

This is great when you're debugging or working on a new feature and want to see the server communication, so it probably served well in the plugin's early life, but I believe it has a real performance cost. Consider times when there is a lot of traffic exchange especially, like debugging: the server sends events for every thread start and thread death. If you're running code with `Future`s, for instance, ensime-vim spends a fair amount of time receiving and logging all these messages that we don't currently even handle.

We should consider a leveled logging approach to improve this, at least two levels, probably the equivalent of `INFO` and `TRACE`/`DEBUG`. Also, it would be nice to clean up and automate messy manual trace logging with decorators.

This might be a good newbie task since it doesn't require much familiarity with ENSIME specifics but you will get to know the code base in the process.

This perhaps could be broken into steps:

1. Move to a persistent `logging.Logger` instance, avoiding all the excess file handle opening. The log sink could be made configurable so e.g. test harnesses can send it to standard out/err (more palatable after trace output is not on by default…), while file remains the default since that's what is sensible when running in Vim.
2. Change trace logging to use an appropriate log level. Ideally try to avoid unnecessary work like pretty-formatting data structures when they're not being logged at higher levels (I don't think we actually explicitly do this now, but just an example). The level should be the `INFO` equivalent by default, with a Vim config variable to change it which we can ask users to set when troubleshooting.
3. Get rid of noisy trace logging statements that pervade the code base using decorators. [Autologging] looks nice, there are other similar solutions if we're willing to accept a pip dependency for this.

If we're willing to use a third-party dep it might be easy to just skip to step 3 and get most of this for free.

[Autologging]: http://pythonhosted.org/Autologging/

ChrisCoffey:
I'd be happy with this landing in two or even a single pull, provided we go the decorator route. 

Autologging does look quite nice. That being said, I think the attractiveness of the package lies in the decorators, and I'm not sure they're worth the cost of requiring users to install a third package (although I could go either way on this since I really don't like reinventing things).

Thoughts on package vs. roll our own?

ches:
[...]

PR #323, opened 2016-07-28
Use a proper logger

ches:
This is a first pass at #297. It does not attempt to hide trace logging code with decorators, etc. yet, I'm punting on the question of adding a dependency or rolling our own for now.

This uses the standard library `logging` module to get rid of the hacky inefficiency of our own homegrown `log` function that re-opens a file handle on every single call. It also adds sensible leveled logging, putting our trace-level logging statements at `DEBUG` level and logging more quietly than that by default.

For now I've just used an environment variable, `ENSIME_VIM_DEBUG`, as a binary toggle between `INFO` and `DEBUG` logging levels. I could add a `g:ensime_debug` Vim variable, we could make it possible to specify where logging goes instead of being hardcoded to `.ensime_cache/ensime-vim.log`, etc., but initially I wanted to add minimal features, just leveled logging with an as-simple-as-possible toggle for the level. Open to feedback.

I'm already finding it helpful to have file and line number of log statements in the output, and the `logging` module's native ability to show exception backtraces in the log. Here's an example of some debug output:

```
DEBUG    <2016-07-28 17:04:03,109> (ensime.py:272) - send: in
DEBUG    <2016-07-28 17:04:03,109> (ensime.py:275) - send: {"callId": 0, "req": {"typehint": "ConnectionInfoReq"}}
DEBUG    <2016-07-28 17:04:04,126> (ensime.py:750) - unqueue: result received
{
  "callId": 0,
  "payload": {
    "typehint": "ConnectionInfo",
    "implementation": {
      "name": "ENSIME"
    },
    "version": "1.0"
  }
}
DEBUG    <2016-07-28 17:04:04,126> (protocol.py:48) - handle_incoming_response: in {u'implementation': {u'name': u'ENSIME'}, u'typehint': u'ConnectionInfo', u'version': u'1.0'}
WARNING  <2016-07-28 17:04:04,126> (protocol.py:61) - Response has not been handled: {u'implementation': {u'name': u'ENSIME'}, u'typehint': u'ConnectionInfo', u'version': u'1.0'}
DEBUG    <2016-07-28 17:04:05,145> (ensime.py:750) - unqueue: result received
{
  "payload": {
    "typehint": "IndexerReadyEvent"
  }
}
DEBUG    <2016-07-28 17:04:05,145> (protocol.py:48) - handle_incoming_response: in {u'typehint': u'IndexerReadyEvent'}
DEBUG    <2016-07-28 17:04:05,145> (ensime.py:386) - message: in
DEBUG    <2016-07-28 17:04:05,145> (ensime.py:387) - [ensime] Indexer is ready
DEBUG    <2016-07-28 17:04:07,177> (ensime.py:750) - unqueue: result received
{
  "payload": {
    "typehint": "SendBackgroundMessageEvent",
    "detail": "Initializing Analyzer. Please wait...",
    "code": 105
  }
}
DEBUG    <2016-07-28 17:04:07,177> (protocol.py:48) - handle_incoming_response: in {u'code': 105, u'typehint': u'SendBackgroundMessageEvent', u'detail': u'Initializing Analyzer. Please wait...'}
WARNING  <2016-07-28 17:04:07,177> (protocol.py:61) - Response has not been handled: {u'code': 105, u'typehint': u'SendBackgroundMessageEvent', u'detail': u'Initializing Analyzer. Please wait...'}
DEBUG    <2016-07-28 17:04:08,187> (ensime.py:750) - unqueue: result received
{
  "payload": {
    "typehint": "AnalyzerReadyEvent"
  }
}
DEBUG    <2016-07-28 17:04:08,187> (protocol.py:48) - handle_incoming_response: in {u'typehint': u'AnalyzerReadyEvent'}
DEBUG    <2016-07-28 17:04:08,187> (ensime.py:386) - message: in
DEBUG    <2016-07-28 17:04:08,187> (ensime.py:387) - [ensime] Analyzer is ready
DEBUG    <2016-07-28 17:04:08,187> (ensime.py:750) - unqueue: result received
{
  "payload": {
    "typehint": "FullTypeCheckCompleteEvent"
  }
}
DEBUG    <2016-07-28 17:04:08,187> (protocol.py:48) - handle_incoming_response: in {u'typehint': u'FullTypeCheckCompleteEvent'}
```

### DIFF
--- a/ensime_shared/config.py
+++ b/ensime_shared/config.py
@@ -17,9 +17,9 @@ gconfig = {
     "localhost": "http://127.0.0.1:{}/{}",
 }
 
+# Messages for user feedback, possible l10n fodder. Please keep alphabetized.
 feedback = {
     "analyzer_ready": "Analyzer is ready",
-    "displayed_type": "The type %s has been displayed",
     "failed_refactoring": "The refactoring could not be applied (more info at logs)",
     "full_types_enabled_off": "Qualified type display disabled",
     "full_types_enabled_on": "Qualified type display enabled",
@@ -36,7 +36,6 @@ feedback = {
     "spawned_browser": "Opened tab {}",
     "start_message": "Server has been started...",
     "typechecking": "Typechecking...",
-    "unhandled_response": "Response has not been handled: %s",
     "unknown_symbol": "Symbol not found",
 }
 
--- a/ensime_shared/config.py
+++ b/ensime_shared/config.py
@@ -10,6 +10,8 @@ from ensime_shared.util import Util
 BOOTSTRAPS_ROOT = os.path.join(os.environ['HOME'], '.config/ensime-vim/')
 """Default directory where ENSIME server bootstrap projects will be created."""
 
+LOG_FORMAT = '%(levelname)-8s <%(asctime)s> (%(filename)s:%(lineno)d) - %(message)s'
+
 gconfig = {
     "ensime_server": "ws://127.0.0.1:{}/jerky",
     "localhost": "http://127.0.0.1:{}/{}",
@@ -17,7 +19,7 @@ gconfig = {
 
 feedback = {
     "analyzer_ready": "Analyzer is ready",
-    "displayed_type": "The type {} has been displayed",
+    "displayed_type": "The type %s has been displayed",
     "failed_refactoring": "The refactoring could not be applied (more info at logs)",
     "full_types_enabled_off": "Qualified type display disabled",
     "full_types_enabled_on": "Qualified type display enabled",
@@ -34,7 +36,7 @@ feedback = {
     "spawned_browser": "Opened tab {}",
     "start_message": "Server has been started...",
     "typechecking": "Typechecking...",
-    "unhandled_response": "Response {} has not been handled",
+    "unhandled_response": "Response has not been handled: %s",
     "unknown_symbol": "Symbol not found",
 }
 
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -1,8 +1,8 @@
 # coding: utf-8
 
-import datetime
 import inspect
 import json
+import logging
 import os
 import shutil
 import sys
@@ -11,13 +11,13 @@ import time
 from subprocess import PIPE, Popen
 from threading import Thread
 
-from ensime_shared.config import commands, feedback, gconfig
-from ensime_shared.debugger import DebuggerClient
-from ensime_shared.errors import InvalidJavaPathError
-from ensime_shared.launcher import EnsimeLauncher
-from ensime_shared.protocol import ProtocolHandler, ProtocolHandlerV1, ProtocolHandlerV2
-from ensime_shared.typecheck import TypecheckHandler
-from ensime_shared.util import catch, module_exists, Util
+from .config import commands, feedback, gconfig, LOG_FORMAT
+from .debugger import DebuggerClient
+from .errors import InvalidJavaPathError
+from .launcher import EnsimeLauncher
+from .protocol import ProtocolHandler, ProtocolHandlerV1, ProtocolHandlerV2
+from .typecheck import TypecheckHandler
+from .util import catch, module_exists, Util
 
 # Queue depends on python version
 if sys.version_info > (3, 0):
@@ -49,10 +49,9 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
     which stores the a handler per response type.
     """
 
-    def __init__(self, vim, launcher, config_path):  # noqa: C901 FIXME
+    def __init__(self, vim, launcher):  # noqa: C901 FIXME
         def setup_vim():
             """Set up vim and execute global commands."""
-            self.vim = vim
             if not int(self.vim_eval("exists_enerrorstyle")):
                 self.vim_command("set_enerrorstyle")
             self.vim_command("highlight_enerror")
@@ -61,25 +60,38 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
             self.vim.command(
                 "autocmd FileType package_info nnoremap <buffer> <Space> :call EnPackageDecl()<CR>")
             self.vim.command("autocmd FileType package_info setlocal splitright")
-            super(EnsimeClient, self).__init__()
 
-        def setup_logger_and_paths():
-            """Set up paths and logger."""
-            osp = os.path
-            self.config_path = osp.abspath(config_path)
-            config_dirname = osp.dirname(self.config_path)
-            self.ensime_cache = osp.join(config_dirname, ".ensime_cache")
-            self.log_dir = self.ensime_cache
-            if not osp.isdir(self.ensime_cache):
+        # Our use case of a logger per class instance with independent log files
+        # requires a bunch of manual programmatic config :-/
+        def setup_logger():
+            path = os.path
+            config = self.launcher.config
+            projectdir = path.abspath(config['root-dir'])
+            project = config.get('name', path.basename(projectdir))
+            logger = logging.getLogger(__name__).getChild(project)
+
+            if os.environ.get('ENSIME_VIM_DEBUG'):
+                logger.setLevel(logging.DEBUG)
+            else:
+                logger.setLevel(logging.INFO)
+
+            # The launcher also creates this - if we refactored the weird
+            # lazy_initialize_ensime path this could go away. Fixes C901.
+            logdir = config['cache-dir']
+            if not path.isdir(logdir):
                 try:
-                    os.mkdir(self.ensime_cache)
+                    os.mkdir(logdir)
                 except OSError:
-                    self.log_dir = "/tmp/"
-            self.log_file = os.path.join(self.log_dir, "ensime-vim.log")
-            with open(self.log_file, "w") as f:
-                now = datetime.datetime.now()
-                tm = now.strftime("%Y-%m-%d %H:%M:%S.%f")
-                f.write("{}: {} - {}\n".format(tm, "Initializing project", config_dirname))
+                    logger.addHandler(logging.NullHandler())
+                    return logger
+
+            logfile = path.join(logdir, 'ensime-vim.log')
+            handler = logging.FileHandler(logfile, mode='w')
+            handler.setFormatter(logging.Formatter(LOG_FORMAT))
+
+            logger.addHandler(handler)
+            logger.info('Initializing project - %s', projectdir)
+            return logger
 
         def fetch_runtime_paths():
             """Fetch all the runtime paths of ensime-vim plugin."""
@@ -93,13 +105,16 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
 
             return paths
 
-        setup_logger_and_paths()
+        super(EnsimeClient, self).__init__()
+        self.vim = vim
+        self.launcher = launcher
+
+        self.log = setup_logger()
+        self.log.debug('__init__: in')
         setup_vim()
-        self.log("__init__: in")
 
         self.ws = None
         self.ensime = None
-        self.launcher = launcher
         self.ensime_server = None
 
         self.call_id = 0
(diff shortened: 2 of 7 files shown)


ktonga:
Love it, logs are actually a big PITA when I have to hack on the plugin. This will improve the contributing experience tremendously.

ensime_shared/config.py
```diff
@@ -5,14 +5,16 @@
...
+
 gconfig = {
     "ensime_server": "ws://127.0.0.1:{}/jerky",
     "localhost": "http://127.0.0.1:{}/{}",
 }
 
 feedback = {
     "analyzer_ready": "Analyzer is ready",
-    "displayed_type": "The type {} has been displayed",
+    "displayed_type": "The type %s has been displayed",
```
ktonga: Unrelated, but why are this kind of messages defined here if they are not feedback to the user? They might be moved to the log statement itself.

ensime_shared/protocol.py
```diff
@@ -232,10 +232,10 @@ def handle_doc_uri(self, call_id, payload):
 
     def handle_completion_info_list(self, call_id, payload):
         """Handler for a completion response."""
+        self.log.debug('handle_completion_info_list: in')
         completions = payload["completions"]
-        self.log("handle_completion_info_list: in")
         self.suggestions = [completion_to_suggest(c) for c in completions]
-        self.log("handle_completion_info_list: {}".format(self.suggestions))
+        self.log.debug('handle_completion_info_list: %s', self.suggestions)
```
ktonga: This kind of output is useless. We could use `pprint.pformat` here and in similar places.

ensime_shared/config.py
ches: I agree it's pretty annoying to need to cross-reference another file if it's strictly a *log* message.

ensime_shared/protocol.py
ches: I think you're referring generally to data structure output in cases where it's a large structure.

The advantage with the old-style `%s` format strings is that the `logging` module elides the actual string formatting if the log level isn't active, so CPU cycles aren't needlessly spent. The `str` representations are fine in many cases like the sample output above IMO, but for larger messages like this where it may be warranted the optimal thing to do is something like this:

```python
if self.log.isEnabledFor(logging.DEBUG):
    self.log.debug('handle_completion_info_list:\n%s', pprint.pformat(self.suggestions))
```

Slightly more intrusive code, but worth it if the output is otherwise pretty useless.

I'm happy to update cases like this in the PR if you all help point them out, I haven't taken the time to consider/check the output of each case too closely, beyond a quick assessment of what level is appropriate.

ktonga: Yeah, I'm aware of the cost of the string generation. A possible solution could be to wrap data-like log args with  something like this:

```python
class DataFmt(object):
    def __init__(self, data):
        self.data = data

    def __str__(self):
        return pprint.pformat(self.data)

self.log.debug("blah %s", new DataFMT(dictionary))
```

Remember, I'm not a python dev, so it might seem a little too JVMish :)

ches: Indeed that could cut down on the conditional statements. Whether creating and GCing those wrapper objects is significantly more expensive than (possibly cached) calls to `log.isEnabledFor` and worth the tradeoff for readability is up for debate, but we're really in the weeds of micro-optimization at that point 😁   I don't feel strongly either way.

ches:
This one's next in the queue for me whenever you have time to look it over @ChrisCoffey.

Does anyone see other cases that would benefit from pretty-printing? If it's just the completions @ktonga pointed out above I'll just put the `pprint` under a simple `if` statement for now, if there are more I can try the wrapper class idea to minimize the distraction.

ensime_shared/ensime.py
```diff
@@ -738,7 +747,7 @@ def trigger_callbacks(_json):
                 now = time.time()
             else:
                 result = self.queue.get(False)
-                self.log("unqueue: result received {}".format(str(result)))
+                self.log.debug('unqueue: result received\n%s', result)
```
ktonga: I'm not quite sure what type `result` is, but this looks like a good candidate for pretty printing as well.

ensime_shared/protocol.py
```diff
@@ -45,18 +45,20 @@ def register_responses_handlers(self):
 
     def handle_incoming_response(self, call_id, payload):
         """Get a registered handler for a given response and execute it."""
-        self.log("handle_incoming_response: in {}".format(payload))
+        self.log.debug('handle_incoming_response: in %s', payload)
```
ktonga: pretty print as payload is a dict

ensime_shared/protocol.py
```diff
@@ -45,18 +45,20 @@ def register_responses_handlers(self):
...
         def feature_not_supported(m):
             msg = feedback["handler_not_implemented"]
             self.raw_message(msg.format(typehint, self.launcher.ensime_version))
+
         if handler:
             with catch(NotImplementedError, feature_not_supported):
                 handler(call_id, payload)
         else:
-            self.log(feedback["unhandled_response"].format(payload))
+            self.log.warning(feedback['unhandled_response'], payload)
```
ktonga: pretty print as payload is a dict

ensime_shared/ensime.py
ches: These are the JSON responses, which somehow are already formatted.

ensime_shared/protocol.py
```diff
@@ -150,7 +152,8 @@ def add(member, indentLevel):
 
     def handle_symbol_search(self, call_id, payload):
         """Handler for symbol search results"""
-        self.log(payload)
+        self.log.debug('handle_symbol_search: in %s', payload)
```
ktonga: pretty print as payload is a dict

ches:
Okay, I hadn't initially thought of pretty-print *all* dicts but I guess it makes sense in `DEBUG`, otherwise how do you really know when one is going to be "small" or not…

I'll make these updates later today, thanks for pointing them out.

ensime_shared/protocol.py
ches: This one is actually called like 10 lines after the `unqueue` log message in the comment just above that outputs the raw JSON, which is apparently already formatted. So they're usually consecutive log lines and it looks pretty redundant and noisy to me (even for debug logging) to have both. I think we're pretty certain at this stage that there is no bug in our call to `json.loads` 😄 

So what seems sensible to me then is a one-line message here, showing `call_id` instead of the payload, to make sense of things if you're analyzing a few messages arriving concurrently. I guess it can also include `typehint` for server-initiated messages with no call ID.

ches:
I've added pretty-printing of the dicts suggested as well as for `send_request`. Also inlined the log messages that were in the `feedback` dictionary that we agreed was annoying.

ensime_shared/util.py
```diff
@@ -56,3 +57,17 @@ def module_exists(module_name):
...
+    """Wrapper to pretty-format object's string representation.
+
+    Reduces boilerplate for logging statements where we don't want to eagerly
+    :func:`pprint.pformat` when the logging level isn't enabled.
+    """
+
+    def __init__(self, data):
+        self._data = data
+
+    def __str__(self):
```
ktonga: Can we get rid of the ugly `u'...'` output implementing `__unicode__` instead?

```
[{'abbr': u'config()',
  'dup': 1,
  'menu': u'SocketChannelConfig',
  'word': u'config'},
```

ches: Doesn't seem to work, that's probably happening within `pprint` for understandable reasons of accuracy. It should go away in Python 3… 😁

ktonga:
The log file isn't truncated anymore, I like it. But does it has any rolling/cleanup policy? Or it should be done manually?

ches:
It should be truncated on each restart as it was before, the write mode on the line

```python
handler = logging.FileHandler(logfile, mode='w')
```

does this. Seems to be working that way for me?

There are time- and size-based rotation handlers in the standard lib that could be used, but I'd rather that decision be addressed separately, I was aiming to preserve the existing behavior.

ktonga:
>  I was aiming to preserve the existing behavior.

That's why I'm asking, it's not behaving the same for me. I exit Vim (the only instance) and enter again, and all the previous log lines are still there. It's not too important if we do something about it later.

ches:
That's weird, I can't reproduce. `head -n 1 .ensime_cache/ensime-vim.log` shows me a different timestamp after each restart when I'm running this branch.

ktonga:
For me the log file is accumulating the lines from multiple Vim sessions. I'll delete it to start fresh and try again.

ches:
Maybe `lsof | grep 'ensime-vim.log'` just as another insurance check something unexpected isn't going on?

ches:
Also `make clean` to clear `pyc` files just as another sanity check…

ensime_shared/ensime.py
```diff
@@ -202,9 +210,9 @@ def setup(self, quiet=False, bootstrap_server=False):
         def lazy_initialize_ensime():
             if not self.ensime:
                 called_by = inspect.stack()[4][3]
-                self.log(str(inspect.stack()))
-                self.log("setup(quiet={}, bootstrap_server={}) called by {}()"
-                         .format(quiet, bootstrap_server, called_by))
+                self.log.debug(str(inspect.stack()))
```
ktonga: The actual output isn't that useful with this format (single line):

```
DEBUG    <2016-08-02 01:09:28,127> (ensime.py:213) - [(<frame object at 0x7f61bb4ef450>, '/home/gaston/Projects/ensime-vim/ensime_shared/ensime.py', 213, 'lazy_initialize_ensime', ['                self.log.debug(str(inspect
.stack()))\n'], 0), (<frame object at 0x7f61bb50f638>, '/home/gaston/Projects/ensime-vim/ensime_shared/ensime.py', 239, 'setup', ['        return self.running and lazy_initialize_ensime() and ready_to_connect()\n'], 0), (<frame object at 0x7f61bb9bea50>, '/home/gaston/Projects/ensime-vim/ensime_shared/ensime.py', 981, 'client_for', ['            if client.setup(quiet=quiet, bootstrap_server=bootstrap_server):\n'], 0), (<frame object at 0x15fa860>, '/home/gaston/Projects/ensime-vim/ensime_shared/ensime.py', 956, 'current_client', ['                create_client=create_client)\n'], 0), (<frame object at 0x16042c0>, '/home/gaston/Projects/ensime-vim/ensime_shared/ensime.py', 878, 'wrapper2', ['                create_client=create_client)\n'], 0), (<frame object at 0x7f61bb985810>, '/home/gaston/Projects/ensime-vim/rplugin/python/ensime.py', 171, 'au_cursor_moved', ['        super(NeovimEnsime, self).au_cursor_moved(*args, **kwargs)\n'], 0), (<frame object at 0x7f61bb9bce18>, '/home/gaston/.local/lib/python2.7/site-packages/neovim/plugin/host.py', 74, '_on_request', ['        rv = handler(*args)\n'], 0), (<frame object at 0x7f61bb9c2d70>, '/home/gaston/.local/lib/python2.7/site-packages/neovim/api/nvim.py', 151, 'filter_request_cb', ['            result = request_cb(self._from_nvim(name), args)\n'], 0), (<frame object at 0x7f61bb9be850>, '/home/gaston/.local/lib/python2.7/site-packages/neovim/msgpack_rpc/session.py', 185, 'handler', ['                rv = self._request_cb(name, args)\n'], 0)]
```

ktonga:
Working now after deleting the file. Maybe related to the old file not being created by the logging framework. Weird.

ktonga:
LGTM

ensime_shared/ensime.py
ches: True, I hadn't really looked at this one yet.

I think the entire fact that this logging statement exists is a smell, related to #325—trying to trace how the launcher actually runs is really convoluted, which I guess is why this is here. There may be some good reason for the way it's done but I don't understand what it is yet. I'd like to push it up to happening from the `Ensime` class without `EnsimeClient` ever touching the launcher directly, but of course that's some separate work.

Is it worth doing anything with this right now?

ktonga: Nope. Go ahead and merge it.

ches:
Thanks for the review and testing @ktonga!

ktonga:
You're welcome! That's nothing compared to the excellent work you're doing to improve the codebase!!

Issue #321, opened 2016-07-27
Factor Vim calls into an Editor object abstraction


ches:
This idea seemed to be supported here: https://github.com/ensime/ensime-vim/pull/314#discussion_r71131986, so this issue is to track a to-do for implementing it. ~~This is potentially low-hanging fruit for a new contributor—though it would seem to touch a lot of code, it should be straightforward and will give a good orientation.~~ Okay, I've started it, and this probably doesn't qualify as low-hanging fruit…

There is currently a `commands` dictionary in `ensime_shared/config.py` containing Vim commands and expressions that are executed throughout the codebase. The dictionary was used with the noble intention of making Vimscript functionality more reusable and maintainable, but in practice it is seeming that real methods on an object would achieve that more successfully. The dictionary results in pervasive code like this:

```python
self.vim.command(commands['some_key_name_for_a_command'].format('some parameters'))
```

This is awkward, doesn't succeed in reducing stringly-typed programming, and gets confusingly mixed with `EnsimeClient.vim_command` which can't deal with passing parameters. It would be nice if e.g. this:

```python
commands = {
    # ...
    "split_window": "split {}",
    "vert_split_window": "vsplit {}",
    "new_vertical_window": "{}vnew {}",
    "new_vertical_scratch":
        "{}vnew {} | setlocal nobuflisted buftype=nofile bufhidden=wipe noswapfile",
    # ...
}
```

became something like this:

```python
class Editor(object):
    def __init__(self, vim):
        self._vim = vim

    def split_window(self, file, vertical=False, size=None, bufopts=[]):
        self._vim.command('...')  # etc. TODO
```

This would make code easier to write, since tools can auto-complete and show argument hints for function calls instead of constantly referencing `config.py`, and easier to test since the tightly-coupled `vim` object could now be easily injected/mocked.

PR #248, opened 2016-04-30
Introduce lettuce tests

ChrisCoffey:
This is the very beginnings of addressing #136. Here's an overview of the changes:

 - Removed all broken tests
 - Introduced Lettuce for Cucumber-style BDD tests
 - Implemented testing for config file parsing
 - Wired up to Drone

While Lettuce seems to be a fairly mature library, I'm worried that it isn't being that actively maintained (last commit Jan 2016). For what we're doing right now I think it's fine, but if we start running up against limitations we may want to switch to Aloe, an active fork.

I'm not convinced Lettuce will meet all of our test needs, so I'm open to adding in additional frameworks that fill any missing gaps.

I'm not a Python developer, so if anyone has advice on the correct way to do things here please jump right in & let me know.

PR #327, opened 2016-07-29
Add unit testing framework, pytest

ches:
I've been doing a bit of work on a project to extract a core ENSIME client for Python that could potentially be shared with ensime-sublime or others (more of a rewrite of the core than extraction, but I'll save this whole discussion until it's really ready to talk about).

I set up unit test infrastructure for that project from the start and thought about whether to add it to ensime-vim too, but wasn't sure if there was a need yet. Now that I've started some work on #321, it's finally there—we're finally getting to the point where some substantial logic will be decoupled enough from complex dependencies to feasibly be unit tested. I've [mentioned before how I think Lettuce is ideal for integration/acceptance/end-to-end tests][1] but unit tests are a better fit for other cases.

An example is the config file parsing, which we had Lettuce tests for but they feel heavy because there's no need for integration there, it's extra work to write steps in code plus a prose feature for something so low-level. So I used that as a small contained example case for the initial unit test framework setup, converting those to unit tests and taking the opportunity to abstract it in a cute little mapping class that I've used in my spinoff project.

Briefly, a rundown of popular unit testing choices in the Python ecosystem—there are three main contenders:

1. The standard library `unittest` module.
2. [Nose](http://nose.readthedocs.io/en/latest/).
3. [pytest](http://docs.pytest.org/en/latest/).

`unittest` has the advantage of being built-in, but it's very much jUnit-inspired and ceremonious: everything is a class, and the worst to me is that you need to learn a bunch of different `assert*` methods.

Nose and pytest by contrast share a lot in common, they both aim to be lightweight frameworks that are highly-plugin driven and let you just write simple functions, scaling up to using more complicated structures if you feel the need to. I chose Pytest for a few reasons:

- Nose 1 development has basically stopped, they're working on [Nose 2][2] which seems mostly ready but the ecosystem is still catching up.
- With nose you either use a plugin to add assertion functions, use `unittest` for assertions with nose acting as a test runner, or use the built-in `assert` statement with not-very-good failure output, kind of like with Lettuce where there is more burden on you to write string explanations for assertions. pytest [has badass output with simple standard assert statements][3].
- I know there was concern about Lettuce not being actively maintained, and AFAIK not ready for Python 3. If we eventually need to leave it behind, [pytest_bdd][4] looks like a really nice Gherkin feature runner based on pytest, maybe even already more featureful and better-documented than Lettuce and working on Python 2/3.

A couple of things that might be negatives for pytest: it is somewhat "magical": its fantastic output for standard `assert` is achieved through bytecode manipulation. That said, it's been around for many years and is regarded as solid. Also, it has [a novel approach to setup/teardown or test fixtures][5] that is like dependency injection—it's actually quite nice and scales up and down very flexibly, but it takes a bit of getting used to since it is unusual. As the doc mentions, there is still support for classical setup/teardown functions at different scopes, though.

Let me know what you think. I'd ideally like to base my work for #321 atop this test setup as well as the logging in #323.

[1]: https://github.com/ensime/ensime-vim/pull/248#issuecomment-224980603
[2]: http://nose2.readthedocs.io/en/latest/index.html
[3]: http://docs.pytest.org/en/latest/example/reportingdemo.html
[4]: https://github.com/pytest-dev/pytest-bdd
[5]: http://docs.pytest.org/en/latest/fixture.html


ktonga:
I think this is great, I totally agree with your __Lettuce vs Unit Tests__ analysis about when each kind of test is better.
When I had to write the tests for the `symbol_format` module it felt like a lot of work, maintain two separate files and so many extra steps to transform the data from what Gherkin supports to what the logic actually needs.

I'm more used to libs like `specs2` where you have BDD-like DSL to organise your cases and all the power of the PL to do your testing. I think that having the cases descriptions in a separate files is a better fit for projects where the acceptance stage is managed by a separate team, most of the time non-developers, but I think it's a big overhead for a project like this with just a group of maintainers.

Having the alternative to eventually move to Python3 is a huge win also.

If this change is accepted I'd be happy to migrate the Suggestion Format scenario to unit tests.
And let me tell you that all the work you're doing is fantastic, it's much more enjoyable to contribute to a project well structured and with the right set of tools.

It's a big LGTM

ches:
> I'm more used to libs like `specs2` where you have BDD-like DSL to organise your cases and all the power of the PL to do your testing.

I hear you, coming from a bunch of Ruby and Scala development over the past many years those kinds of test frameworks have spoiled me, one of the things I miss most dearly doing Python. so_many_underscores_to_write_tests. The language just isn't optimized to support DSLs that way. Attempts like [NoseOfYeti][1] (what a name…) and [Konira][2] never catch on because they're crazy preprocessor-like hacks…

> And let me tell you that all the work you're doing is fantastic, it's much more enjoyable to contribute to a project well structured and with the right set of tools.

Recently I've had some time to devote to the project, it probably won't last forever, so for myself too I feel a rush to try to whip the foundations into shape while I can so it's more comfortable to build upon… Thanks for the kind words!

[1]: http://noseofyeti.readthedocs.io/en/latest/examples.html
[2]: https://github.com/alfredodeza/konira

ches:
Bonus: added coverage.py, with the coverage report combining the pytest and Lettuce suites.

ches:
A little trivial update: moved the Lettuce tests into `test/features`, figured the consolidated layout would be easier for contributors to understand. That was basically just a `git mv` and change the path in the `Makefile`. Also fits with conventions many tools expect, e.g. no longer needed to tell `coverage.py` to explicitly ignore the `*steps.py` files as not being production code.

If we port the suggestion formatting tests to pytest also, there won't be any more Lettuce tests remaining for now, but #312 could change that…

PR #328, opened 2016-07-31
Editor Object (#321)

ches:
~~This branch is based on my logger and unit testing branches, #323 and #327. It will be a lot easier to review if those get merged. In the meantime here is the diff as if they were already merged:~~

~~https://github.com/ches/ensime-vim/compare/integration...ches:editor-object~~

This is a first pass at factoring all editor functionality out of the `EnsimeClient` class and into a separate one, as described in #321. I would still like to make further improvements to the API of the `Editor` object, but since this is already a lot of changes, I've tried to mostly make this phase a "direct port" of existing functions for the benefit of easier review. The follow-up changes might evoke more design discussion, hopefully this two-phase approach helps to defer most of that to when it easier to focus on.

The `config.commands` dictionary is completely eliminated 🎉 

The `vim` object is eliminated from `EnsimeClient` with a couple of exceptions that I'll note inline.

New API on the `Editor` object is covered by unit tests mocking the `vim` object. I haven't written tests for the old API yet because I want to change a lot of it in the next stage. By the way, `make coverage` is a thing now.

I would appreciate it if brave souls can try running this branch for their work for a few days (cc @ChrisCoffey @ktonga). This is a lot of changes on top of two already significant branches, so I would really like to start landing some of it—I think altogether they are big steps in making ensime-vim easier for contributors to dive into.


ensime_shared/ensime.py
```diff
@@ -49,41 +50,46 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
...
-
-        def setup_logger_and_paths():
-            """Set up paths and logger."""
-            osp = os.path
-            self.config_path = osp.abspath(config_path)
-            config_dirname = osp.dirname(self.config_path)
-            self.ensime_cache = osp.join(config_dirname, ".ensime_cache")
-            self.log_dir = self.ensime_cache
-            if not osp.isdir(self.ensime_cache):
+    def __init__(self, editor, vim, launcher):  # noqa: C901
```
ches: I can't *quite* jettison the `vim` attribute yet, the remaining usage is for `disable_plugin` which is too complicated to deal with in the course of these changes, I've dived into it a bit for #294 and will continue later.

ensime_shared/ensime.py
```diff
@@ -465,16 +369,16 @@ def send_at_point_req(self, what, path, row, col, size, where="range"):
...
 
     def type_check_cmd(self, args, range=None):
         """Sets the flag to begin buffering typecheck notes & clears any
         stale notes before requesting a typecheck from the server"""
-        self.log("type_check_cmd: in")
+        self.log.debug('type_check_cmd: in')
         self.start_typechecking()
         self.type_check("")
-        self.vim.command('silent !echo "Typechecking..."')
+        self.editor.message('typechecking')
```
ches: This `!echo` was basically bug—pointlessly executing on the shell, usually leaving Vim needing a redraw. I'm surprised no one has reported this yet, it wouldn't be a bad idea to hotfix this ahead of the other changes.

ensime_shared/protocol.py
```diff
@@ -197,54 +193,52 @@ def handle_string_response(self, call_id, payload):
...
+                    self.log.exception('handle_doc_uri: webbrowser error')
+                    self.editor.raw_message(feedback["manual_doc"].format(url))
 
             del self.call_options[call_id]
         else:
+            # TODO: Get rid of this self.vim usage, but I don't understand
+            # when/if we ever get here...
             self.vim.current.buffer[:] = \
                 [line.encode('utf-8') for line in payload["text"].split("\n")]
             self.en_format_source_id = None
```
ches: If anyone knows what this `else` branch is supposed to do, I'd appreciate being enlightened, because this is the only other remaining `self.vim` aside from `disable_plugin`.

`:EnDocBrowse` works on this branch, `:EnDocUri` is broken on master AFAICT and I've never really known what it's supposed to do/what the point of it is anyway. `:EnDocUri` doesn't hit this code path either, it fails in the `if` branch. No idea what `self.en_format_source_id` is supposed to have to do with this.

ensime_shared/ensime.py
ches: Oh, I also changed this in the course of the logging PR #323 that should go through before this anyway.

ensime_shared/ensime.py
```diff
@@ -944,9 +812,7 @@ def teardown(self):
 
     def current_client(self, quiet, bootstrap_server, create_client):
         """Return the current client for a given project."""
-        # Use current_file command because we cannot access self.vim
-        current_file_cmd = commands["current_file"]
-        current_file = self.vim.eval(current_file_cmd)
+        current_file = self.vim.current.buffer.name
```
ches: I didn't understand the comment here because it still uses `self.vim` anyway 😩   This change seems to work fine, but let me know if anyone sees an issue.

The cases I believe we know of where using the `vim` object can be problematic:

- Off of the main thread. The Python Vim API isn't threadsafe so that's why the only code that tries to invoke it from the polling worker thread, which is `disable_plugin`, jumps through the `threadsafe_vim` hoop (which is just YOLO for standard Vim…).
- In `Ensime.__init__`. We observed that using `self.vim` there didn't work and my hunch was that's because the class is first being defined when an autocommand like `VimEnter` fires triggering it to be loaded, and we're also *defining the code that Vim is trying to invoke for our autocommand handler* within the `Ensime` class, so there's a reentrancy thing going on there.

Neither of these apply here. It may have been that at one time, further back than I have spelunked in `git blame`, that `current_client` was being called from `__init__`?

ensime_shared/protocol.py
```diff
@@ -138,16 +131,18 @@ def add(member, indentLevel):
...
-        cmd = commands["new_vertical_scratch"].format(str(45), "package_info")
-        self.vim.command(cmd)
-        self.vim.command(commands["set_filetype"].format("package_info"))
-        self.vim.command(commands["append_line"].format("\'$\'", str(package)))
+        opts = {'buftype': 'nofile', 'bufhidden': 'wipe',
+                'buflisted': False, 'swapfile': False}
+        self.editor.split_window('package_info', vertical=True, size=45, bufopts=opts)
+        self.editor.set_filetype('package_info')
+
+        self.editor.append(str(package))
```
ches: It'd be nice to have an even higher-level API like `Editor.open_package_inspector()` or something like that so that message handler code could think in terms of "UI components" and not low-level Vim buffer options, but that's the sort of stuff I want to save for a later review/design session.

ensime_shared/protocol.py
```diff
@@ -160,37 +155,34 @@ def handle_symbol_search(self, call_id, payload):
...
-                                             p["line"],
-                                             str(sym["name"]),
-                                             "info")
+                item = self.editor.to_quickfix_item(str(p["file"]),
+                                                    p["line"],
+                                                    str(sym["name"]),
+                                                    "info")
                 qfList.append(item)
-        self.write_quickfix_list(qfList)
+        self.editor.write_quickfix_list(qfList)
```
ches: Ditto on the higher-level API.

ensime_shared/protocol.py
ktonga: What I understand after a quick read of the source code:

This method `handle_doc_uri`, indirectly through `handle_string_response` is used to handle all responses of type `StringResponse` (very tricky, good candidate to be refactored #330)

Currently there is two very different kind of message handle by this method:

- __DocBrowse__: handled in the `True` branch of the `if` where all the _open-with-browser-magic_ is done
- __FormatSource__: handled in the `else` branch under question, you get here with the __:EnFormatSource__ command which sends the message that is responded with a `StringResponse` containing the formatted source code. This command also sets `en_format_source_id` to a __True-like__ value in [ensime.py:397](https://github.com/ches/ensime-vim/blob/8db528a32e7466b7a9df2326d75764d4c8dc4085/ensime_shared/ensime.py#L397). The text of the `current.buffer` is replaced with the one from the response using the `vim` instance. So you could move this to the `Editor Object` as something like `editor.replaceText(payload['text'])` which I think is really dangerours since the request can take a few seconds to be responded and the current buffer could potentially be a different one. IMO this feature should be safely implemented or removed.

ches: Ah! Thanks for the second set of eyes. I hadn't traced through further up the stack that way. That is totally bonkers.

I hope that if people use `:EnFormatSource` they give it a moment to work before making more changes to the buffer, but better UI feedback there could at least help in the interim if we can't come up with a better solution. I suppose that's a matter for a new day though.

ches: Don't know how I missed it, right there one function above in the source and even with a comment 😩  Still, bonkers.

ktonga:
Hey @ches `:EnSuggestImport` isn't working. Didn't do any further research tho. Sorry.
From the logs I can tell you that the request reaches the server but nothing happens afterwards. No errors are shown in the logs.

ches:
Hmm that's weird, it works for me. Let me know if you have a chance to figure out more specific reproduction conditions.

ktonga:
Never mind, it seemed to be an unrelated problem which I solved just restarting the server. After commenting here I started noticing problems with other commands, such as `:EnType`, so I decided to restart and that fixed all the problems. Maybe it's related somehow but I would not be able to reproduce it easily. I will let you know if it happens again. Sorry for the noise.

ensime_shared/editor.py
```diff
@@ -0,0 +1,384 @@
...
+        buf = self._vim.buffers[bufnr] if bufnr else self._vim.current.buffer
+
+        for opt, value in options.items():
+            buf.options[opt] = value
+
+    # TODO: make this a R/W property?
+    def set_filetype(self, filetype):
+        """Set filetype for the current buffer.
+
+        Unfortunately setting the filetype with ``set_buffer_options`` (or
```
ktonga: This is an implementation detail, no need to be leaked in the editor API. Maybe it could be specially handled in `set_buffer_options` instead of being a separate method.

ches: That's probably true, I guess I thought this might be independently useful, but it could also be automatically handled as a special case in `set_buffer_options`. There's no other singular form of that anywhere just because it wasn't needed yet.

ensime_shared/editor.py
```diff
@@ -0,0 +1,384 @@
...
+        return {"filename": file_name,
+                "lnum": line_number,
+                "text": message,
+                "type": tpe}
+
+    def write_quickfix_list(self, qflist):
+        self._vim.command('call setqflist({!s})'.format(qflist))
+        self._vim.command('copen')
+
+    def lazy_display_error(self, filename):
```
ktonga: Is it for displaying compilation errors? Isn't this functionality provided by Syntastic already?

ensime_shared/editor.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def clean_errors(self):
+        """Clean errors and unhighlight them in vim."""
+        self._vim.eval('clearmatches()')
+        self._errors = []
+        self._matches = []
+        # Reset Syntastic notes - TODO: bufdo?
+        self._vim.current.buffer.vars['ensime_scala_notes'] = []
+
+    def message(self, key):
```
ktonga: Can't we combine `message` and `raw_message` into a single method `message(key, args)` so we don't need to do the formatting outside? We could also encapsulate the use of the `feedback` dict. Maybe suitable for the second round of changes you are planning.

ensime_shared/ensime.py
ktonga: This should be delegated on a method in `editor` considering we ain't gonna reference `vim` directly from the client anymore.

ensime_shared/editor.py
ches: Yeah most definitely want to clean this up, but it falls under "direct port" for this review. I know I've created some new API already but I believe it's all stuff that wasn't encapsulated in a function already.

ensime_shared/editor.py
ches: This is, I think, for built-in error highlighting support that ensime-vim has if Syntastic is not installed. I briefly tried it and I'm not sure if it's even correctly working at this point, but nevertheless just left the code intact for now.

I left a TODO comment also related to this where we have a setting that is supposed to be to let the user customize the highlight group used for this built-in error highlighting, I think, but it looks like the code doesn't actually use the value of the setting…

ensime_shared/ensime.py
ches: This is a method of the `Ensime` class, not `EnsimeClient`—we can discuss this more in the context of design discussion in the second part of these changes, but my reasoning so far is that using `vim` here is reasonable because the `Ensime` class basically represents *the Vim plugin itself* on the Python side, i.e. it's the base class for `NeovimEnsime` for instance.

ktonga:
This change is looking so sexy. I've been using it for a week or so and seems to be working well.

LGTM

You should rebase on `master` to integrate the new debug commands and you can merge it.
Excellent work!

ensime_shared/ensime.py
ktonga: Sorry, my bad for looking at the decontextualized diff. I agree with you on the scope of the `vim` object.

ches:
Thanks for the review and trying it out!

I'll incorporate a change of the `set_filetype` consolidation thing, and I'd also maybe like to make an interim improvement on #330 before rebasing this. Maybe not the bigger changes discussed there, but at least breaking the source formatting part into a sensible separate function should be easy and now I understand what it's doing in order to get rid of that last `self.vim` usage. Been preoccupied with other stuff for the last few days so haven't gotten around to that yet, should be able to in the next or two.

ktonga:
Ok. Sounds great. I'll keep using it for a few more days so, just to be sure. But be aware that I only use a small subset of the plugin functionality.

ches:
Okay, I've rebased this and incorporated changes we've discussed.

PR #329, opened 2016-07-31
Add MAINTAINERS and lgtm.co config

ches:
This sets up [LGTM](https://lgtm.co/docs/overview/), which Sam already enabled for this repo, so we need approval to merge **this** branch now 😄   It make master a protected branch requiring a "LGTM", +1, 👍 , or :shipit: on PRs to merge.

Please let me know if you would like anything here changed.

I don't know what's up with GitHub rendering in the diff but it's mangling "Gastón" in my browser even though the file is properly encoded and looks correct on my machine. I'm assuming it will work, but will fix if not…

@jvican @ChrisCoffey @ktonga

### DIFF
--- a/.lgtm
+++ b/.lgtm
@@ -1,0 +1,2 @@
+approvals = 1
+pattern = "(?i):shipit:|:\\+1:|\\+1|LGTM"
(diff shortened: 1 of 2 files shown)


ktonga:
It's working perfect on my phone, so seems to be a desktop-only problem. Maybe we can send them a PR, oh wait! @github is not on GitHub.

Unicode is a new concept and that "diff" feature isn't used that much anyway.

ktonga:
At the risk of sounding snooty... it LGTM

ktonga:
Hey @ches MAINTAINERS is always pulled from `master` so we are having a chicken and egg situation in this PR.

ches:
I wondered if that was going to happen… seems like an obvious gotcha situation that lgtm.co needs to solve. But we might just need to have @fommil toggle the integration off until we merge this then.

PR #284, opened 2016-06-17
[WIP] Integrate debugger stepping (#212)

Jacoby6000:
Addresses #212

There's still some more debugger stuff that can be done, like updating values, and probably highlighting the line currently suspended.  There could also probably be something we can do for showing where existing breakpoints are, and maybe listing symbols in a quickfix menu or something.

I'm still new to vim plugin development, so I'd like help figuring those parts out!

Please let me know if there's anything I need to change.

PR #331, opened 2016-08-01
Integrate debugger stepping. Closes #212

### LINKED ISSUE #212: Implement Debug Stepping
Right now we've implemented the ensime  server's breakpoint api, but not the step api. While you can accomplish stepping with just breakpoints, we may as well support the full step api.

ktonga:
This is a cleanup of what @Jacoby6000 started in #284 
It needs to land before code keeps diverging after so many refactors

Add support for:

- Step
- Step-Out
- Next


ches:
Needs a `make format` 😄

ktonga:
Yeap. Done.

ensime_shared/debugger.py
```diff
@@ -65,3 +65,21 @@ def debug_backtrace(self, args, range=None):
...
+            "threadId": self.debug_thread_id})
+
+    def debug_step_out(self, args, range=None):
+        self.log("debug_step_out: in")
+        self.send_request({
+            "typehint": "DebugStepOutReq",
+            "threadId": self.debug_thread_id})
+
+    def debug_next(self, args, range=None):
+        self.log("debug_next: in")
```
ches: Since the logging is merged these will need to be `self.log.debug` not to break.

ktonga: Ups, you're right.

ches:
There we go, LGTM!

Issue #330, opened 2016-08-01
Revisit `handle_string_response` implementation. The current one is very confusing.


ktonga:
I think the handling around text responses is currently hard to follow, it handles very different kind of messages and the code branching is base on non-clearly-named global state.

ches:
A saner approach than `call_options` and the way class attributes are sometimes set as flags in some far-off location in the code (which is also dependent on sequential processing of the queue) probably boils down to a fundamental design change that I'm not sure I personally will even try to rework before working instead on my side project to rewrite the client core—I'm attempting to use an event loop instead of a brute-polling thread, seems like the efficient choice for something I/O-bound.

`handle_string_response` can at least dispatch to separate functions with appropriate names for handling the different requests that it determines with the state variables, in the shorter term.

Maybe there could be some kind of small interim improvement by maintaining a shared dictionary of `call_id` -> `typehint`, so an overloaded response message type like `StringResponse` could look up more precisely what the request type was instead those icky flag attributes?

ktonga:
Yes, that would be enough for a short term solution. And maybe keep a dict `call_id` -> `buffer name` for source formatting to always replace the content in the right buffer.

```python
self.editor.replaceText(buffer_names[call_id], payload['text'])
buffer_names.pop(call_id, None)
```

ches:
Ah that's what you meant about changing buffer, I had thought you meant editing the current buffer's content. That's a good point.

So mechanically this stuff is basically what `call_options` already is. We could piggyback on that, possibly rename it to something that reflects the expanded significance.

ktonga:
Oh I wasn't aware of `call_options` it's so global that it is almost invisible. A good approach to give it more visibility would be to add it to the handler's params.

```python
    def handle_incoming_response(self, call_id, payload):
        ...
        # make it safer (handle missing entries)
        options = self.call_options[call_id]
        handler(call_id, payload, options)
        ....

    # then in some handler implementation
    def handle_symbol_info(self, call_id, payload, options):
        ...
        display = options.get("display")
        ...

```

PR #332, opened 2016-08-15
Make StringResponse handler's roles more obvious

ches:
This is an incremental refactoring in response to #330 and the discussion that spawned it: https://github.com/ensime/ensime-vim/pull/328#discussion_r72899574

This does not attempt to solve some broader design issues discussed in #330, it’s just a small localized change to separate the logic for the very different types of requests that `StringResponse` can be returned for, so it's more obvious that this handler has these different roles. This is in part to help factor out the direct usage of the editor to finish #328. I've made these private functions for now, they're not really handler functions for discrete ENSIME server response message types like the rest of the formal `ProtocolHandler` interface. They will might eventually fit logically somewhere else.

As I've mentioned elsewhere, I don't really understand the purpose of the `:EnDocUri` command, it used to print the documentation URL in Vim's message area—which I don't see as useful—but even that is broken on current master (the command actually causes an exception). My guess is that the intent was that it could be useful as a *function* instead of command, allowing users to do something custom with the URL as a function result. As-is I don't yet see how to cleanly change it to work that way, it would have to be made synchronous. So with this change, I have un-documented the command but left some implementation in place with a TODO (and it no longer causes an exception, just basically a no-op). I could just go all the way and completely remove the command and that handling code right now, WDYT?

### DIFF
--- a/doc/ensime.txt
+++ b/doc/ensime.txt
@@ -143,15 +143,6 @@ available.
     works offline using cached doc jars. See |ensime-custom-browser| to
     customize the browser used if heuristics don't pick the one you prefer.
 
-                                                                   *:EnDocUri*
-:EnDocUri
-
-    Displays documentation URL for the symbol under the cursor.
->
-    TODO: seems broken for me, and like :EnSymbol I can't think of good use
-    cases to support exposing this as a command, perhaps a function.
-<
-
                                                              *:EnFormatSource*
 :EnFormatSource
 
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -588,7 +588,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
         """Browse doc of whatever at cursor."""
         self.log.debug('browse: in')
         self.call_options[self.call_id] = {"browse": True}
-        self.doc_uri(args, range=None)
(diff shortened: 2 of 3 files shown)


ktonga:
LGTM

ktonga:
Let's merge it so you can finish with #328

ches:
Thanks, I guess I need to LGTM myself since I don't know if anyone else is actively reviewing these days 😄 

About to step AFK, should be able to wrap up #328 tomorrow.

ches:
LGTM

ktonga:
I thought it was configured to need just one LGTM

ktonga:
The check seems to be not working

ches:
I thought so too… I'm not going to like this thing if it's broken…

ktonga:
Re: Doc URI:

I think, if it is possible, it would be nice to add an optional arg to `:EnDoc` which should act as action. Missing arg means `browse` (open in browser) and any provided value should be the name of a function that takes a string (the resultant URI) as parameter and performs whatever action is desired. `:EnDocUri` should be removed.

Open doc as usual:
`:EnDoc`

Perform custom action with URI:
`:EnDoc ShowMessage`

Somewhere in the `vimrc`:
```VimL
function! ShowMessage(msg)
    echom a:msg
endfunction
```

ktonga:
LGTM

Issue #325, opened 2016-07-29
Factor launcher instance out of EnsimeClient entirely


ches:
I'm on a mission to, piece by piece, strip `EnsimeClient`'s responsibilities down to connecting to a server, sending it requests, and dispatching the responses to handlers. This is up for debate, but I think its attempts to manage its own server process lifecycle should be extracted.

`EnsimeClient` takes an instance of `EnsimeLauncher` as a constructor param. IMHO abstraction-wise, and testing-wise, the client object shouldn't really have anything to do with a launcher instance directly: the server should be launched, then that job is done and a client can be instantiated (it's useful to instead pass only the parsed `.ensime` project config—`launcher.config`—to the client constructor, and that makes sense to me abstraction-wise).

Within `EnsimeClient`, the launcher instance is essentially only used in the `setup()` method, which is called constantly on cursorhold (!!) with weird and confusing parameters. In a convoluted way, this serves as a process monitor, potentially re-launching and reconnecting if the server dies. I feel that all of this can and should be moved out to either the `Ensime` class or some kind of separate watchdog class that it instantiates. It can just call a `reconnect()` method on the client when needed, or possibly tell `Ensime` to create an entirely fresh client instance, whatever makes sense.

Goodbye, launcher and ENSIME process handle attributes of `EnsimeClient`.

This most likely ends up having some substantial overlap with #294. I've got a little bit of WIP on that, but I've got a little bit of WIP on a lot of things so I don't want it to deter anyone else from picking up these tickets, just ping me in case I can help give you a head start.

Issue #243, opened 2016-04-27
Remove redundant fun_ prefix


ChrisCoffey:
Remove `fun_` prefix from functions introduced by the plugin. Purely a stylistic point, but it would be nice to clean up the names.

yyadavalli:
I can send a PR for this, if we are moving forward with the change

jvican:
Hey @yyadavalli, I think that @ches was planning to hack on this. There're some opened issues that need to be resolved, though. Maybe you would be interested in picking them, especially the ones related to the debugger. This would be a great feature to have (it's not complete atm).

yyadavalli:
@jvican  Sure, I am looking into the debugger related issues, not sure how much I can do on those. I will comment on those once I am satisfied that there is some progress

ches:
@ChrisCoffey @jvican You can assign this to me, I'm working on it.

I'm experimenting with some refactoring for the plugin initialization—I would do this in (at least) two pull requests with just a minimal renaming per subject of this issue coming first, but I do want to try it. You know how much I've complained about all the boilerplate of defining duplicate shims for Vim and Neovim, which has proved error-prone several times already, plus the devolution into defining basically all of the functions and commands with the most general interface possible that conveys no meaning to users and feels kinda like we're doing stringly-typed programming or using `varargs: Any*` absolutely everywhere…

So what I'm hoping is that I can make a first pass that can eliminate all of the redundant command and function definitions of `NeovimEnsime`, *without* needing to touch anything to do with the `@execute_with_client` decorator (yet, at least). I think this is probably achievable by using the nvim RPC APIs which there is already a `TODO` for in `call_plugin`, or something close to that. I'm digging into those now.

Every single one of the commands and functions is currently `sync=True` anyway, though of course I want to leave open a clean path to selectively take advantage of async on the Neovim side in the future where it's possible. Let's see what it ends up looking like.

ChrisCoffey:
Very interesting, I'm looking forward to seeing what this looks like.

Issue #244, opened 2016-04-27
Clean up Function Signatures for plugin-specific functions


ChrisCoffey:
Right now `@execute_with_client` takes two parameters, `args` & `range`. These make a great deal of sense for commands, but it's more confusing for functions introduced for other workflows like omnicomplete or package exploration.

jvican:
I do not mind to have this feature if we have a general way to bind between vim functions and python functions without increasing the code size. The whole purpose of `@execute_with_client` is to remove all the unnecessary code.

Issue #64, opened 2015-10-02
Provide generic API to ensime for other Vim plugins to integrate.


kstep:
I wasn't sure where should I start this discussion, so I decided to open an issue here.

The problem with current ensime-vim is it's trying to everything. E.g. it tries to highlight errors, but Syntastic plugin already dues it very well, much more better, than ensime-vim will be able to do in near future. I started an experiment in my fork (setloclist branch in kstep/ensime-vim), and now I have Syntastic and ensime play really well together!

The same thing is with YCM, which makes autocompletion very well (though it can integrate with ensime, if only it set omnifunc instead of completefunc). The same thing with Unite (I can imagine imports suggestions and symbols search with Unite, should be really awesome!).

I hope you got my point. The problem with ensime-vim, as I see it now, is it tries to play alone, doing things others already do much more better, than ensime-vim itself.

The one way to go out of this pit, as I see it, to make ensime-vim to provide generic Vim API to ensime calls, and then let community create integration plugins with other general-purpose plugins, like Syntastic, Unite etc.

I personally ready to provide Syntastic integration for errors highlighting, which already work for me very well.

yazgoo:
Good idea.
Could you please provide an example of vim API.
Keeping in mind that ensime is asynchronous, so we should at least allow to register callback for specifics events.

todesking:
I strongly agreed :+1: :+1: 
"Do one thing well" plugin is better than huge monolithic ones.

Maybe we need like this:
```vim
function IntegrateToSyntastic(path, errors)
  " ...
endfunction

call ensime#register_note_event(function('IntegrateToSyntastic'))
```

kstep:
@todesking yes, that's one thing to do it! I thought about something more general, though, like:

```vim
call ensime#on_signal("NewErrors", function("AppendErrors"))
call ensime#on_signal("ErrorsComplete", function("ShowErrors"))
call ensime#emit_signal("GetErrors")
```

With possible helper functions like `fun ensime#on_errors()` and `fun ensime#get_errors()` which wrap more general forms.

I think I can do it, it can be simple enough refactoring from my Syntastic integration implementation.

Ir1sh:
I really like this idea and would just like to make one comment regarding it.  How does providing an api for other existing plugins to interact with effect the 'onboarding' process for new people to either vim or ensime vim?  having to install ensime and ensime-vim and then go find each separate plugin that is going to interface with ensime-vim might be a pain.

[...]

PR #335, opened 2016-08-26
Begin Ensime class refactoring

ches:
This is a bit of preliminary work eventually leading toward #325, #294, and #243. The big contiguous parts of the diff are just moving `EnsimeClient` to its own file, there are no changes to that code otherwise so you can mostly ignore those.

We’ve had some recent discussion about how responsibilities should be delineated between the `Ensime` and `EnsimeClient` classes and possibly other helpers. Following from that, `Ensime` has some methods that don’t fit or at least should not be public API, etc. This takes a few initial steps to clean up.

I've begun work on getting the server launcher and its process management out of `EnsimeClient`'s direct responsibility, in part to simplify the convoluted initialization path that goes like this:

```
execute_with_client decorator ->
  current_client ->
    client_for ->
      create_client ->
        EnsimeClient.setup
```

with several arguments that are strung all along the way. This has bugged me for awhile. But combined with the changes on this PR it's starting to look hairy to review, so it seemed best to push these more modest changes initially and leave the ones touching more complex logic for a focused subsequent review.


autoload/ensime.vim
```diff
@@ -3,10 +3,6 @@ if !has('nvim')
     execute ':redraw!'
 endif
 
-function! ensime#client_keys() abort
-    return s:call_plugin('client_keys', [])
-endfunction
```
ches: This is a breaking change to the Vim scripting API—I'm making the bold assumption that no one is actually using this function… what it returns is a list of absolute paths to `.ensime` files for possibly multiple projects open in the Vim session. This could potentially be useful, but if so, the name sucks anyway—it's totally an implementation detail that the config file paths are a "key" used for caching client instances internally.

I'd like to put off commitment to stability of our Vimscript API for later; it is ugly right now and should change dramatically whenever we take the time to give its design due consideration (related: #243, #244, #64).

ensime_shared/ensime.py
```diff
@@ -771,32 +33,36 @@ class Ensime(object):
...
 
-    def init_settings(self):
-        """Loads all the settings from the ``g:ensime_*`` namespace.
-
-        Invoked on client creation to avoid ``autocmd`` deadlocks.
-        """
-        self.server_v2 = bool(self.get_setting('server_v2', 0))
+    def using_server_v2(self):
+        """Whether user has configured the plugin to use ENSIME v2 protocol."""
+        return bool(self.get_setting('server_v2', 0))
```
ches: After giving it thought (prompted by pylint which raises an eyebrow at instance attributes defined outside of `__init__`), I thought an accessor-type function here was nicer at least while there's only one setting. It avoids the question of where/when `init_settings` gets called due to the matter of `vim` use in the constructor, and it could also allow for switching dynamically between server versions without restarting Vim, which could be useful for development especially after we add some server lifecycle management.

ktonga:
LGTM

I've been using this branch today and I haven't found any problem. I few comments came to my mind while reviewing but they were out of the scope that you very well defined.

BTW, thanks for the stepped PR'ing. Really helpful for reviewing.

ches:
Thanks for the review. Feel free to comment if you have thoughts that are related, I'm getting pretty heavily into changes to take all server lifecycle stuff out of `EnsimeClient` now and there's a lot that I'm undecided about, just seeing how some ideas feel as I go along. It might be awhile before I have time to finish it up. I guess if I get so a decent checkpoint where things work but aren't complete I could open an early PR for feedback.

Still want to get back to the `Editor` API considerations too, but I felt like these things are more important and they aggravate me more 😁

PR #336, opened 2016-09-01
Better server installation path

ktonga:
The installation folder has now the pattern `{scala_minor}/{ensime_version}`.

Examples:

- `2.10/1.0.0`
- `2.11/2.0.0-SNAPSHOT`

It has two benefits over the previous folder name (Scala full version):

- Ensime plugin will not ask for `:EnInstall` for different
  Scala patch versions. Ex: (2.11.7 vs 2.11.8)
- Easy switch between Ensime Server V1 and V2

### DIFF
--- a/ensime_shared/launcher.py
+++ b/ensime_shared/launcher.py
@@ -65,9 +65,11 @@ class EnsimeLauncher(object):
         self.ensime_version = self.ENSIME_V2 if server_v2 else self.ENSIME_V1
         self._config_path = os.path.abspath(config_path)
         self.config = ProjectConfig(self._config_path)
+        self.scala_minor = self.config['scala-version'][:4]
         self.base_dir = os.path.abspath(base_dir)
         self.classpath_file = os.path.join(self.base_dir,
-                                           self.config['scala-version'],
+                                           self.scala_minor,
+                                           self.ensime_version,
                                            'classpath')
         self._migrate_legacy_bootstrap_location()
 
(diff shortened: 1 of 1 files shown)


ches:
I'm on board with incorporating the server version to accommodate switching during the transition period—I believe that was an acknowledged todo when we started structuring code for the protocol split—but I mused at one point in Gitter about whether we could only install the server per Scala minor version and for some reason that I didn't fully understand or don't recall, @fommil had said it was necessary to match the patch version. Maybe I misunderstood or something has changed?

It's pretty much bikeshedding, but a subdirectory seems a little tidier to me than underscores.

Also, we probably ought to provide an auto-migration like the one that already exists for yet another legacy installation path…

ktonga:
The assembly jar is only one per minor (for all patches)  so I don't see a reason why it should be different for sbt artifacts resolution. I recall @fommil saying that only the minor has to match, but I challenge  you to find anything in the new version of gitter.
I don't have any strong fillings about the sub-folder vs underscore thing, so if you prefer the former I will change it.

fommil:
Minor version should be fine. It used to need patch but we fixed that... although I'm sure there will be runtime exceptions for some of the really early releases in each line. Way i look at it, if somebody wants 2.11.0 support they have the freedom and instructions to compile it themselves.

ches:
> Way i look at it, if somebody wants 2.11.0 support they have the freedom and instructions to compile it themselves.

I concur.

ches:
I had missed the updated patch here—LGTM 👍

ches:
Oh wait—@ktonga would you mind looking at providing automatic migration like the existing `_migrate_legacy_bootstrap_location()` does?

ktonga:
Ups, merged. It's so easy, sorry. I'll take a look at that tonight.

ktonga:
I don't get how the `_migrate_blah...` is migrating anything. The server is first downloaded to the new location and this method is just moving the old dir from top-level `.config` dir into ensime-vim specific one. But in this new case the old location is already in the same root folder.

What I could do in this case is, if I detect a server in the previous location, I could relocate it to the new path but I should run some heuristic on the files to detect the server version. Honestly I think that could cause some problems so my  take is to let the users just run `:EnInstall` to get the server installed in the right folder. If they updated the plugin and checked the changelog they might expect it.

Let me know what you think. I'm ok doing whatever is best.

PR #341, opened 2016-09-26
Fix regression in non-Syntastic typecheck error highlighting

### LINKED ISSUE #339: Error highlighting doesn't work anymore
Error highlighting stopped worked recently in Neovim, I can't really say when but around one month ago I think, vim still sees the errors coming from the server, and they are reported in Vim in the message area on the bottom, but they are not highlighted.

This is my configuration: 

Neovim: NVIM 0.1.3-dev Build type: Dev  

I tried with sbt-ensime plugin version: 1.0.0 and 1.9.0 and 1.9.1-SNAPSHOT
all have the same problem.

Thanks, 
Luigi

ches:
Regression from #328 where I didn't properly cast Vim's string `'0'` result from the Syntastic feature detection to Boolean, and escaping was doubled when changing highlight match pattern to a Python raw string.

I believe this fixes the issue—the highlighting only works for me in GUI, not on console, but I *think* that's something with my setup or something that wasn't changed by #328 anyway. This feature is still kinda janky altogether, I made notes in #328 about the logic for the highlighting group customization seemingly not making sense, etc.

It'd be good to add a test or two but it would be a bit hairy to do, so I'll save that until the further `Editor` API improvements, and maybe we can give this feature a little love overall.

Closes #339. @gigiigig you might be happy with Syntastic now, but if you're able to try this with it disabled and confirm if it restores the behavior you had before, that'd be awesome.

### DIFF
--- a/ensime_shared/editor.py
+++ b/ensime_shared/editor.py
@@ -357,7 +357,7 @@ class Editor(object):
         """Renders "notes" reported by ENSIME, such as typecheck errors."""
 
         # TODO: this can probably be a cached property like isneovim
-        hassyntastic = bool(self._vim.eval('exists(":SyntasticCheck")'))
+        hassyntastic = bool(int(self._vim.eval('exists(":SyntasticCheck")')))
 
         if hassyntastic:
             self.__display_notes_with_syntastic(notes)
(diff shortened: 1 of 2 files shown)


gigiigig:
@ches I tried this and it works for me, thanks for the quick fix!

ktonga:
Working for me as well.

LGTM

PR #345, opened 2016-10-22
Remove :EnFormatSource

### LINKED ISSUE #344: Remove `EnFormatSource` command
Support for formatting source will be removed from server soonish (ensime/ensime-server#1606), so the removal of the command will be mandatory when talking to a server 2.0.
Since defining different commands per server version sounds so painful I see two possible options:
- Totally remove the command
- Overwrite the command handler for 2.0 as NOOP with message `Not supported in 2.0`

yyadavalli:
Follow up of https://github.com/ensime/ensime-server/issues/1606

Resolve #344

### DIFF
--- a/ensime_shared/protocol.py
+++ b/ensime_shared/protocol.py
@@ -188,15 +188,9 @@ class ProtocolHandlerV1(ProtocolHandler):
         This is the response for the following requests:
           1. `DocUriAtPointReq` or `DocUriForSymbolReq`
           2. `DebugToStringReq`
-          3. `FormatOneSourceReq`
         """
         self.log.debug('handle_string_response: in [typehint: %s, call ID: %s]',
                        payload['typehint'], call_id)
-
-        if self.en_format_source_id:  # User requested :EnFormatSource
-            self._format_source_file(payload['text'])
-            self.en_format_source_id = None
-            return
 
         # :EnDocBrowse or :EnDocUri
         url = payload['text']
@@ -212,11 +206,6 @@ class ProtocolHandlerV1(ProtocolHandler):
             # TODO: make this return value of a Vim function synchronously, how?
             self.log.debug('EnDocUri %s', url)
             return url
-
-    def _format_source_file(self, newtext):
-        formatted = [line.encode('utf-8') for line in newtext.split('\n')]
-        # FIXME: should assure original buffer, not whatever is now current
-        self.editor.replace_buffer_contents(formatted)
 
     def _browse_doc(self, url):
         self.log.debug('_browse_doc: %s', url)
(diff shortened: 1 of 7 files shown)


ensime_shared/protocol.py
```diff
@@ -213,11 +207,6 @@ def handle_string_response(self, call_id, payload):
             self.log.debug('EnDocUri %s', url)
             return url
 
-    def _format_source_file(self, newtext):
-        formatted = [line.encode('utf-8') for line in newtext.split('\n')]
-        # FIXME: should assure original buffer, not whatever is now current
-        self.editor.replace_buffer_contents(formatted)
```
ktonga: Please remove `replace_buffer_contents` from `Editor` class since this was its only usage and we try to avoid having dead code.

ches:
I never personally use the ENSIME source formatting, because projects using Scalariform to enforce standards generally had sbt-scalariform formatting upon compile anyway (and extra effort had to be made to get ENSIME to use the same settings).

That said though, I don't know how many others may be using it—should we be removing this before server v2 is released?

yyadavalli:
Yes, or you ever support calling sbt tasks from vim, it should call `ensimeScalariformOnly`. The api endpoint is remove in server v2

ktonga:
@yyadavalli sorry I just saw you have already pushed the changes suggested in the reviews.
@ches could you approve please as he has also achieved your suggestion?

ches:
I'm not sure my point was understood here: I realize that formatting is removed in server v2, *but server v2 hasn't shipped yet*. So we've removed functionality that people may *currently* be using for no real reason.

It's done now, since I don't personally use it I don't really mind, but if some users complain that they were using this feature it might warrant reverting.

By the way, we should have also removed documentation of the command from `doc/ensime.txt`.

ktonga:
@ches re: server v2 not shipped yet, [it's actually been shipped already](https://github.com/ensime/ensime-vim/issues/300#issuecomment-230166239) and [the change was already merged](ensime/ensime-server/pull/1619). Also it will soon gain more interest since it will be the only way to use Ensime with Scala 2.12.

I don't think we should revert if someone complains, it's the direction taken on the server so sooner or later it has to affect all the editors, so better we think on alternatives more than keeping something almost dead. Don't you think?

On the other hand you're totally right about removing the docs. It's a shame we haven't spotted it before.

ches:
Okay, "shipped" in the sense that we aren't using it yet in ensime-vim's bootstrap installation, then.

Discouraging people from using it since it is going away: I would agree, but IMO it would have been nice if we gave some sort of deprecation warning for the command until we start installing server v2, instead of just ripping it out.

> On the other hand you're totally right about removing the docs. It's a shame we haven't spotted it before.

I've got a few minor doc edits sitting around, I'll make that change too and open a PR with them.

ktonga:
Sorry, I've already opened the PR to remove the docs, I did it before seeing your comment.
I kind of agree on the deprecation cycle, but maybe for a more popular feature :)

I have to work on making server v2 work again, I think it's broken after a few breaking changes. After doing that I'll propose a proper versioning for the server as all the changes are rolled under 2.0.0-SNAPSHOT and wont be a nice user experience if it breaks every so often.

Issue #346, opened 2016-11-07
name 'ensime_plugin' is not defined on Windows (Troubleshooting hints did not help)


paultol:
I installed Ensime plugin to sbt and ensime-vim plugin (using Vim-Plug) on Windows 8.1, vim 8.0.69. 

I have a simple sbt project and ensimeConfig seems to have worked fine.

> ensimeConfig
[info] ENSIME update.
[warn] SBT is using ivy to resolve dependencies which is known to be slow. Coursier is recommended: http://get-coursier.io
[info] Resolving org.scala-lang#scalap;2.11.4 ...
[info] ENSIME processing hello (Hello SBT)
[warn] Some source directories do not exist and will be ignored by ENSIME.
[warn] For example: C:\Users\Pavel\prj\sbt\hello\src\test\scala

In gvim, I receive these error messages every time I try to load .scala file or run any ensime ex command (the commands seem to be available, typing :En and pressing <TAB> lets me scroll through the commands):

Error detected while processing function ensime#com_en_declaration[1]..<SNR>39_call_plugin
line 15:
E605: Exception not caught: <type 'exceptions.NameError'>:name 'ensime_plugin' is not defined

I followed Troubleshooting hints and found that Python versions used by vim and pip modules sexpdata, websocket-client match (they both are 2.7):

results of
:py import sys; print(sys.path):
['C:\\must>not&exist', 'C:\\WINDOWS\\SYSTEM32\\python27.zip', 'C:\\pkg\\python27\\Lib', 'C:\\pkg\\python27\\DLLs', 'C:\\pkg\\python27\\Lib\\lib-tk', 'C:\\Users\\Pavel\\prj\\sbt\\hello', 'C:\\pkg\\vim\\vim80', 'C:\\pkg\\python27', 'C:\\pkg\\python27\\lib\\site-packages', '_vim_path_']

C:\Users\Pavel\prj\sbt\hello>pip show sexpdata | findstr Location
Location: c:\pkg\python27\lib\site-packages
C:\Users\Pavel\prj\sbt\hello>pip show websocket-client | findstr Location
Location: c:\pkg\python27\lib\site-packages

Please help

paultol:
[...]

PR #348, opened 2016-11-10
fixed check for file path ends

paultol:
pull request as agreed at https://github.com/ensime/ensime-vim/issues/346

### DIFF
--- a/autoload/ensime.vim.py
+++ b/autoload/ensime.vim.py
@@ -9,10 +9,18 @@ import vim
 
 def ensime_init_path():
     path = os.path.abspath(inspect.getfile(inspect.currentframe()))
-    if path.endswith('/rplugin/python/ensime.py'):  # nvim rplugin
+    dir_file1 = os.path.split(path)
+    path_fn = dir_file1[1]
+    dir_file2 = os.path.split(dir_file1[0])
+    path_dir1 = dir_file2[1]
+    dir_file3 = os.path.split(dir_file2[0])
+    path_dir2 = dir_file3[1]
+    expected_nvim_path_end = ['rplugin', 'python', 'ensime.py']
+    expected_vim_path_end = ['autoload','ensime.vim.py']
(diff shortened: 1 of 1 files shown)


autoload/ensime.vim.py
```diff
@@ -9,10 +9,18 @@
 
 def ensime_init_path():
     path = os.path.abspath(inspect.getfile(inspect.currentframe()))
-    if path.endswith('/rplugin/python/ensime.py'):  # nvim rplugin
```
ktonga: I think that wrapping the expected path with [`normpath`](https://docs.python.org/2/library/os.path.html#os.path.normpath) should be enough:

```python
if path.endswith(os.path.normpath('rplugin/python/ensime.py')):
```

It will leave it the same for linux and mac but will replace `/` with `\` when on windows

autoload/ensime.vim.py
```diff
@@ -9,10 +9,18 @@
...
+    dir_file2 = os.path.split(dir_file1[0])
+    path_dir1 = dir_file2[1]
+    dir_file3 = os.path.split(dir_file2[0])
+    path_dir2 = dir_file3[1]
+    expected_nvim_path_end = ['rplugin', 'python', 'ensime.py']
+    expected_vim_path_end = ['autoload','ensime.vim.py']
+    if expected_nvim_path_end == [path_dir2, path_dir1, path_fn]: # nvim rplugin
         sys.path.append(os.path.join(
             os.path.dirname(os.path.dirname(os.path.dirname(path)))))
-    elif path.endswith('/autoload/ensime.vim.py'):  # vim plugin
```
ktonga: Same solution here

autoload/ensime.vim.py
paultol: I see your point. normath is doing too many things so I shortened the patch using join

ktonga:
@paultol thank you so much for you contribution!

PR #349, opened 2016-11-14
Remove format source docs

ktonga:


ches:
Included this change in #351, assuming nothing is controversial there I'll close this in lieu of it.

PR #350, opened 2016-11-14
WIP - Add `:EnRestart` command

ktonga:
It's a work in progress only shared to get some early feedback, not ready to be merged yet. Still trying to fix a problem that prevents it to work properly.

There's a few logging lines to debug the problem, please ignore them by now.

### DIFF
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -11,6 +11,10 @@ function! ensime#teardown() abort
     return s:call_plugin('teardown', [])
 endfunction
 
+function! ensime#restart() abort
+    return s:call_plugin('restart', [])
+endfunction
+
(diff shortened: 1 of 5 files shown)


ktonga:
@jvican, yes, i did, sadly it's not working. Something related to the threading system is not working as I expected, as a result it doesn't execute the shutdown sequence of steps in the order the should be executed. That's why I tagged this PR as a WIP.
Was hoping any of the refractors @ches is working on could make the things work.

jvican:
Cool, hope to see this feature soon. In my opinion, it's one of the most important ones. Spurious errors happen in the communication between client and servers and being able to restart the connection is a huge improvement in UX. We also need to figure out what's going on under the hood with the connection. Stability and predictability should be a primer concern for us 🥇 I don't know whether you guys agree on this.

Issue #353, opened 2016-12-03
Launch ENSIME using server jar paths now in .ensime


ches:
Since the 1.12.x series ensime-sbt now installs ENSIME with Coursier and records the server, scala-compiler, etc. jar paths needed for a launcher classpath in `.ensime` (see ensime/ensime-sbt#271).

I've mostly implemented this for ensime-vim today, should finish tomorrow hopefully or will at least open a WIP review. Just wanted to claim the issue to avoid anyone duplicating work.

The emacs client implemented this in ensime/ensime-emacs#563, for reference. I believe #167 was meant to be this but was closed when we added Coursier within the sbt bootstrap—connecting the references for posterity.

PR #354, opened 2016-12-09
Support launching server preinstalled by build tools

ches:
This updates `EnsimeLauncher` to use the server preinstalled by ensime-sbt 1.12+ if available, as per #353. This should be pretty much ready, just want to get feedback/testing and possibly add some unit tests in the meantime now that the launcher doesn't have to do a bunch of slow side-effecty installation stuff that is obnoxious to test or stub.

I've refactored the launcher to use a strategy pattern for launching/installing with three strategies, in order of precedence:

1. **Assembly jar**: for [local development or some corporate behind-the-firewall usage](http://ensime.github.io/contributing/#manual-qa-testing).
2. **.ensime**: use the build tool-installed server if the `.ensime` tells us how.
3. **bootstrap server installation**: the old way, for transition period until people upgrade ensime-sbt and other build tools add install functionality. It'll be more fun to see the future big red diff that removes `SbtBootstrap` 😄 

One change to note: I removed the ability to specify that a bootstrap installation should use server v2. With the new ensime-sbt an `ensimeVersion` setting can be given in an `ensime.sbt` to achieve this much more conveniently. I assume anyone who wants to use server v2 at this point is an ensime-vim developer and you don't mind using sbt and the latest ensime-sbt on your plugin testing project, so it's nice to get rid of this special-case handling.

I think it would be nice to integrate a bit of feedback in the launcher for the user to know what's being used (if their assembly jar is picked up, for instance), either via a logger or in the Vim messages that already announce the server is starting, etc., but IMO that's out of scope for this PR.

### DIFF
--- a/ensime_shared/config.py
+++ b/ensime_shared/config.py
@@ -7,7 +7,7 @@ import sexpdata
 
 from ensime_shared.util import Util
 
-BOOTSTRAPS_ROOT = os.path.join(os.environ['HOME'], '.config/ensime-vim/')
+BOOTSTRAPS_ROOT = os.path.join(os.environ['HOME'], '.config', 'ensime-vim')
 """Default directory where ENSIME server bootstrap projects will be created."""
 
 LOG_FORMAT = '%(levelname)-8s <%(asctime)s> (%(filename)s:%(lineno)d) - %(message)s'
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -111,7 +111,8 @@ class Ensime(object):
         """
         server_v2 = self.using_server_v2()
         editor = Editor(self._vim)
-        launcher = EnsimeLauncher(self._vim, config_path, server_v2)
+        config = ProjectConfig(config_path)
+        launcher = EnsimeLauncher(self._vim, config, server_v2)
         if server_v2:
             return EnsimeClientV2(editor, self._vim, launcher)
         else:
--- a/ensime_shared/ensime.py
+++ b/ensime_shared/ensime.py
@@ -50,8 +50,9 @@ class Ensime(object):
         self._vim = vim
         self.clients = {}
 
+    @property
     def using_server_v2(self):
-        """Whether user has configured the plugin to use ENSIME v2 protocol."""
+        """bool: Whether user has configured the plugin to use ENSIME v2 protocol."""
         return bool(self.get_setting('server_v2', 0))
 
     def get_setting(self, key, default):
@@ -109,11 +110,11 @@ class Ensime(object):
 
         This will launch the ENSIME server for the project as a side effect.
         """
-        server_v2 = self.using_server_v2()
-        editor = Editor(self._vim)
         config = ProjectConfig(config_path)
-        launcher = EnsimeLauncher(self._vim, config, server_v2)
-        if server_v2:
+        editor = Editor(self._vim)
+        launcher = EnsimeLauncher(self._vim, config)
+
+        if self.using_server_v2:
             return EnsimeClientV2(editor, self._vim, launcher)
         else:
             return EnsimeClientV1(editor, self._vim, launcher)
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -193,8 +193,9 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
                 self.log.debug(str(inspect.stack()))
                 self.log.debug('setup(quiet=%s, bootstrap_server=%s) called by %s()',
                                quiet, bootstrap_server, called_by)
-                no_classpath = not os.path.exists(self.launcher.classpath_file)
-                if not bootstrap_server and no_classpath:
+
+                installed = self.launcher.strategy.isinstalled()
+                if not installed and not bootstrap_server:
                     if not quiet:
                         scala = self.launcher.config.get('scala-version')
                         msg = feedback["prompt_server_install"].format(scala_version=scala)
@@ -261,6 +262,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
     def connect_ensime_server(self):
         """Start initial connection with the server."""
         self.log.debug('connect_ensime_server: in')
+        server_v2 = isinstance(self, EnsimeClientV2)
 
         def disable_completely(e):
             if e:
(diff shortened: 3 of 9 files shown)


ensime_shared/config.py
```diff
@@ -7,7 +7,7 @@
 
 from ensime_shared.util import Util
 
-BOOTSTRAPS_ROOT = os.path.join(os.environ['HOME'], '.config/ensime-vim/')
+BOOTSTRAPS_ROOT = os.path.join(os.environ['HOME'], '.config', 'ensime-vim')
```
ktonga: Unrelated, but the config folder location could be improved by using XDG for Linux and AppData for Windows.

ensime_shared/ensime.py
```diff
@@ -48,10 +48,12 @@ def __init__(self, vim):
         # race condition of autocommand handlers being invoked as they're being
         # defined.
         self._vim = vim
+        self._editor = Editor(self._vim)
         self.clients = {}
 
+    @property
     def using_server_v2(self):
-        """Whether user has configured the plugin to use ENSIME v2 protocol."""
+        """bool: Whether user has configured the plugin to use ENSIME v2 protocol."""
```
ktonga: I'm a bit confused, keeping this setting just to choose the client version can't lead to a client and server version mismatch?

ktonga: Nvm, it's an undocumented setting so only people knowing what they do should be using it.

ktonga: Sorry for the noise, I think we should get rid of this setting and ask the launcher which version we should use for the client.

ensime_shared/ensime.py
```diff
@@ -109,13 +111,13 @@ def create_client(self, config_path):
...
         """
-        server_v2 = self.using_server_v2()
-        editor = Editor(self._vim)
-        launcher = EnsimeLauncher(self._vim, config_path, server_v2)
-        if server_v2:
-            return EnsimeClientV2(editor, self._vim, launcher)
+        config = ProjectConfig(config_path)
+        launcher = EnsimeLauncher(self._vim, config)
+
+        if self.using_server_v2:
```
ches: You're right, it isn't ideal to keep a client config setting for this if it's now configurable from the build tools, I hadn't really thought about removing the client setting altogether.

I'm divided about how to go about this, though. Right now the launcher doesn't actually care what version it's launching, and it's kind of nice that it doesn't need to. Parsing it out of jar file names in `:ensime-server-jars` feels icky to me; it's not otherwise directly available as a field in the `.ensime` file, and introducing coupling to particular build tools to get it (e.g. asking sbt for `ensimeVersion`) seems wrong too, at least for this purpose.

IMO what feels right is to get the server version strictly from the handshake messages (`ConnectionInfo` or whatever might possibly come out of ensime/ensime-server#1552 in the future, it appears that hasn't been resolved yet). That's cleanly agnostic to the launcher strategy.

This implies that handshake is completed before an `EnsimeClient` gets instantiated for the appropriate protocol version. That would be a change (subject of a new enhancement, if we agree on this design), but it sounds like a sensible one to me. It would get trickier if this becomes async, but how much so is hard to say since those specifics aren't decided/implemented.

Maybe I'm making too big a deal out of avoiding stuff like parsing version out of the jars… clearly it isn't *hard* to do that, I'm just afraid it leads to multiple launch strategy solutions that are somewhat fragile, and internal coupling that can be avoided—I think there might be conscious reasoning in the fact that server version isn't expressly encoded in `.ensime`, like that a running server should be the authoritative source of truth for protocol version, not `.ensime`, nor the build tool, nor a launcher that's finished its job of opaquely invoking `java` with a classpath.

Thoughts?

If this sounds like the way to go, then I'll create a new ticket for making client protocol selection automatic based on handshake, and `g:ensime_server_v2` will be removed once that gets done.

ktonga: Hmm, you have all valid points, getting the version from the `ConnectionInfo` sound perfect, even more, we need the protocol version no the server version, it's only that this time internal change to the server were the perfect excuse to also change the protocol, but that could not be the case in the future.

The problem I see is that handling websocket messages is currently the main responsibility of the client, so we have a chicken and egg problem here due to the current implementation of the client  relaying on inheritance to pick the protocol version. IMHO the client should be only one, having all the WS machinery and after shacking hands with the server it should instantiate the proper version of protocol handlers.

Go ahead and create the ticket, the change looks big enough to be out of scope of this PR.

ches: Sorry for the delay, couldn't spare time during the week—I created #355 and I'm going to try to wrap this PR up with some tests today.

> the problem I see is that handling websocket messages is currently the main responsibility of the client, so we have a chicken and egg problem here due to the current implementation of the client relying on inheritance to pick the protocol version. IMHO the client should be only one, having all the WS machinery and after shacking hands with the server it should instantiate the proper version of protocol handlers.

Yep, I'm not sure about what's going to work best here yet either. It'd be nice if responsibility for dealing with connection/messages didn't spread, but it still seems to me like mixins are the most manageable way to deal with protocol versions (open to ideas though). From my POV, I think it will be easier to decide what feels right after knocking out #325 and other server lifecycle aspects that will achieve some clean-up refactoring around these parts of the code.

In my separate ensime-py client library effort I began modeling a `Connection` as an explicit object—that could end up being something that *knows* the protocol version for the connection it represents, so maybe that could end up being a clean way for connections to be started and then passed into the constructor of client instances for subsequent communication. That is a long way from being ready or fully thought out, though 😣

ktonga:
Hey @ches I've been trying to give this PR a shot but something is preventing the rplugin from being registered in neovim, I'm on 0.2-dev installed via PPA.
Maybe an exception thrown and not properly handled during startup? Let me know if you cannot reproduce it I can assist you with the diagnosing.

ches:
@ktonga Okay I reproduced that, thanks for the report. It looks like it's similar to the problem before where using the `vim` object within the `Ensime` constructor is problematic, because I put `self._editor = Editor(self._vim)` in the constructor now and internally that does an `eval` to eagerly check `has('nvim')`. Interesting that this didn't seem to be a problem for regular Vim.

I've just pushed an update that reverts that change. Maybe it would be better for avoiding surprises to make `Editor.isneovim` evaluate lazily and memoize instead of doing that evaluation in the constructor—I'll play with that, but for now I'm just reverting to the prior impl.

ches:
I finally got around to adding some initial unit tests that cover at least the new stuff on this PR (updated on the branch now). As we all know the installation part of the sbt bootstrap strategy is unpleasant to unit test so I leave that matter for #312 and/or the day we remove support for it altogether…

I'd like to give `EnsimeProcess` some love eventually too but that's for another day and another ticket.

PR #360, opened 2017-01-01
Fixes #356: Invoking completion after ( replaces the word that comes before

### LINKED ISSUE #356: Invoking completion after ( replaces the word that comes before
### Steps to reproduce

1. Check out https://github.com/jqno/EnsimeVimCompletionIssue
2. Open `CompletionTest.scala`
3. Write `print` and invoke completion; choose `println`.
4. Write `(`
5. Write `he` and invoke completion

### Expected behaviour

The line is expanded to `println(hello)`

### Actual behaviour

The line is expanded to `hello)`

### More info

I tried to fix this bug and offer a PR, but unfortunately I couldn't figure it out. I have noticed that `fun_en_complete_func` is invoked with a `base` parameter of `println(he`, where I would have expected just `he`. However, it seems to come directly from VIM that way, so I'm not sure what the intended approach should be:

* Try to get VIM to send `he` to ensime-vim instead of `println(he`
* Adapt ensime-vim to return `println(hello` instead of just `hello`
* Another approach entirely?

It seems to make sense to combine this with #271. I'd like to take a stab at that too, if you can give me some pointers :).

jqno:
Please see the discussion in issue #356.

### DIFF
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -693,7 +693,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
             row, col = self.editor.cursor()
             start = col
             line = self.editor.getline()
-            while start > 0 and line[start - 1] not in " .":
+            while start > 0 and line[start - 1] not in " .,([{":
                 start -= 1
             # Start should be 1 when startcol is zero
             return row, col, start if start else 1


ktonga:
I'll merge this right away since the fix is really small and it solves a considerable bug.

jvican:
Thank you for improving the quality of this plugin @jqno :tada:!

ches:
Thanks for the fix @jqno, and for taking the trouble of creating a reproduction on the issue ticket!

jqno:
You're welcome! 😃
I maintain an open source project myself, so I know how helpful it is to have a proper reproduction recipe. And your instructions to create one, were quite clear, so thanks for that!

PR #362, opened 2017-01-03
Remove the last raw vim instance from EnsimeClient

ches:
Doing some refactoring, and here's one chunk that stood alone from larger changes.

`disable_plugin` doesn't really work (#294). We might not need it at all if upcoming changes handle problem scenarios better (reconnect to server, throw away client instance for a new one, restart server, etc.), but for now I moved it to `Ensime` with a todo. That's where it probably belongs anyway if fixed since `Ensime` represents the plugin itself, and if there are multiple clients (multiple projects in the Vim session) then they should all get shut down if the plugin is disabled. `Ensime` is currently the place to handle things like that.

### DIFF
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -280,6 +280,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
                 from websocket import create_connection
                 # Use the default timeout (no timeout).
                 options = {"subprotocols": ["jerky"]} if server_v2 else {}
+                options['enable_multithread'] = True
                 self.log.debug("About to connect to %s with options %s",
                                self.ensime_server, options)
                 self.ws = create_connection(self.ensime_server, **options)
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -48,7 +48,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
     which stores the a handler per response type.
     """
 
-    def __init__(self, editor, vim, launcher):  # noqa: C901 FIXME
+    def __init__(self, editor, launcher):  # noqa: C901 FIXME
         # Our use case of a logger per class instance with independent log files
         # requires a bunch of manual programmatic config :-/
         def setup_logger():
@@ -81,24 +81,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
             logger.info('Initializing project - %s', projectdir)
             return logger
 
-        def fetch_runtime_paths():
-            """Fetch all the runtime paths of ensime-vim plugin.
-
-            disable_plugin needs this to run on the main thread, hence the
-            eager execution from constructor.
-            """
-            runtimepath = self.vim.eval('&runtimepath')
-            plugin = "ensime-vim"
-            paths = []
-
-            for path in runtimepath.split(','):
-                if plugin in path:
-                    paths.append(os.path.expanduser(path))
-
-            return paths
-
         super(EnsimeClient, self).__init__()
-        self.vim = vim
         self.editor = editor
         self.launcher = launcher
 
@@ -128,11 +111,6 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
         self.toggle_teardown = True
         self.connection_attempts = 0
         self.tmp_diff_folder = tempfile.mkdtemp(prefix='ensime-vim-diffs')
-
-        # Set the runtime path here in case we need
-        # to disable the plugin. It needs to be done
-        # beforehand since vim.eval is not threadsafe
-        self.runtime_paths = fetch_runtime_paths()
 
         # By default, don't connect to server more than once
         self.number_try_connection = 1
@@ -167,9 +145,9 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
                         connection_alive = False  # noqa: F841
                     else:
                         if not self.number_try_connection:
-                            # Stop everything and disable plugin
+                            # Stop everything.
                             self.teardown()
-                            self.disable_plugin()
+                            self._display_ws_warning()
 
                 # WebSocket exception may happen
                 # FIXME: What Exception class? Don't catch Exception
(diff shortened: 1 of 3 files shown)


ensime_shared/client.py
```diff
@@ -224,26 +202,10 @@ def tell_module_missing(self, name):
...
-
-        for path in self.runtime_paths:
-            self.log.debug(path)
-            threadsafe_vim('set runtimepath-={}'.format(path))
-
+    def _display_ws_warning(self):
         warning = "A WS exception happened, 'ensime-vim' has been disabled. " +\
             "For more information, have a look at the logs in `.ensime_cache`"
-        threadsafe_vim('echo "{}"'.format(warning))
+        self.editor.raw_message(warning)
```
ches: This string should probably go into `feedback` for sake of consistency with most other messaging. I'll save that for later.

ensime_shared/client.py
```diff
@@ -224,26 +202,10 @@ def tell_module_missing(self, name):
...
-    def disable_plugin(self):
-        """Disable plugin temporarily, including also related plugins."""
-        self.log.debug('disable_plugin: in')
-
-        def threadsafe_vim(command):
-            """Threadsafe call if neovim, normal if vim."""
-            def normal_vim(e):
-                self.vim.command(command)
-            with catch(Exception, normal_vim):
-                self.vim.session.threadsafe_call(command)
```
ches: This was actually broken, we couldn't tell because of catching `Exception` 😖 This method takes a function as argument, not a string of a Vim command. `vim.async_call` is a higher-level wrapper for it in the Neovim Python API.

ktonga:
I think the `:UpdateRemotePlugins` thing is happening again.

ches:
Ugh, what the hell. I'm rolling back changes bit by bit and I can't tell what is giving Neovim a problem this time. Turned on nvim's debug logging and not seeing anything that helps me. This is starting to get on my nerves.

ktonga:
Same story here with the server restart PR, super frustrating. Is it better the experience in py3? Would it be hard to migrate for us?

ches:
When I generate a working rplugin manifest by checking out master and then switching back to this branch, I get this in `:messages` at nvim startup:

> Error detected while processing VimEnter Auto commands for "*.scala":
> no request handler registered for "/Users/ches/.vim/bundle/ensime-vim/rplugin/python/ensime.py:autocmd:VimEnter:*.scala"

But we clearly have a `VimEnter` handler registered and that fact hasn't changed—if some error is causing nvim not to load it it'd be nice to get an error message *about that*… and by the way, the plugin still otherwise works once a `CursorHold` or `CursorMove` kicks in!

I comment out all code in the functions that handle `VimEnter` so they're no-op, and also remove all calls on the `vim` object during construction of `Ensime` and `Editor` to rule out that known pitfall, and nothing helps, `:UpdateRemotePlugins` still breaks 😖

The changes on this branch don't even seem consequential to me for the issue that manifests.

ches:
I can't really explain why, but the Neovim issue appears to be fixed with the latest commit.

Our build seems to be broken by recent changes to the Drone infrastructure, see ensime/ensime-emacs#573…

ensime_shared/ensime.py
```diff
@@ -130,13 +130,14 @@ def disable_plugin(self):
...
+        for path in self.runtime_paths():
             self._vim.command('set runtimepath-={}'.format(path))
 
-    @property
+    # Tried making this a @property, with and without memoization, and it made
+    # plugin initialization misbehave in Neovim (only). WTF.
     def runtime_paths(self):  # TODO: memoize
         """All the runtime paths of ensime-vim plugin files."""
-        runtimepath = self._vim.eval('&runtimepath')
+        runtimepath = self._vim.options['runtimepath']
```
ktonga: Maybe the underlying evaluation model for `option` is different from `eval`, since it started working for me as well.

ches: This didn't actually solve it, I tried both, the critical part was the method can't be a `@property`. Still don't know why, because `using_server_v2` has seemed to work fine as a property on this class.

You should see the huge WIP I've finally made progress on that follows after this 😁 No more launcher object in `EnsimeClient` either. Going to need a lot of testing because it's hard to make the changes small, but once it's robust it should hopefully make stuff like the server restarting way easier. Definitely need the build sorted out when that lands…

ktonga: Do you prefer to merge it and fix the build after?

ktonga: I was about to give #350 another shot, but now I'm totally hooked on your WIP :smile:

ches: > Do you prefer to merge it and fix the build after?

I guess it depends on how long it takes to get it sorted out. I'm continuing to work on top of this locally so it doesn't bother me to wait a bit, but this also passes tests for me locally and seems to be working for the two of us, some I'm fairly confident about it.

Here's [a sneak preview][1] of the rest on the way, has a few holes and things to finish but it basically works already!

[1]: https://github.com/ensime/ensime-vim/compare/master...ches:remove-server-lifecycle-from-client

jvican:
I don't remember when but I've had this issue with neovim before. My solution was to reinstall the plugin, but it also struck me that neovim wasn't detecting correctly the hooks after changes to the code. Glad to see that's fixed.

ensime_shared/client.py
jvican: Good catch, this may be related to this error I've seen several times now:
```
error caught while executing async callback:
TypeError("'str' object is not callable",)
TypeError: 'str' object is not callable
 
the call was requested at
  File "/home/jvican/.nvim/plugged/ensime-vim/ensime_shared/util.py", line 51, in catch
    handler(str(e))
  File "/home/jvican/.nvim/plugged/ensime-vim/ensime_shared/client.py", line 171, in logger_and_close
    self.disable_plugin()
  File "/home/jvican/.nvim/plugged/ensime-vim/ensime_shared/client.py", line 236, in disable_plugin
    threadsafe_vim('set runtimepath-={}'.format(path))
  File "/home/jvican/.nvim/plugged/ensime-vim/ensime_shared/client.py", line 232, in threadsafe_vim
    self.vim.session.threadsafe_call(command)
```

ensime_shared/editor.py
```diff
@@ -9,7 +9,7 @@ class Editor(object):
 
     def __init__(self, driver):
         self._vim = driver
-        self._isneovim = bool(self._vim.eval("has('nvim')"))
+        self._isneovim = bool(int(self._vim.eval("has('nvim')")))
```
jvican: Good one 👍

ensime_shared/client.py
jvican: Yes, confirmed that this is definitely gone. :tada:

ches:
Okay, it's a bummer that the new CI environment is messed up but particularly since @jvican has also been able to lend some attention I'd like to capitalize on our momentum and move ahead. I'll remove the commit that updates the build config so that this branch "passes" and can be merged.

ensime_shared/client.py
ches: Oh that's interesting, I'd seen this error in testing but never ran into it in day-to-day usage (I still don't use Neovim 100% of the time). This change seemed correct but I wasn't quite sure because it didn't work as I expected with [REPL testing][1]. Good to know it indeed appears to be working.

[1]: https://github.com/neovim/python-client#usage-through-the-python-repl

PR #367, opened 2017-01-06
WIP: Remove server lifecycle management from EnsimeClient

### LINKED ISSUE #325: Factor launcher instance out of EnsimeClient entirely

ches:
This resolves #325, that ticket serves as a good statement of the purpose of this PR.

I still have a few known todos here and I'm working on writing some tests, but this is such a large and core change (and frankly one that I'm really eager to get out of the way) that I wanted to open it up and get some people trying it. It basically works for me now but I'd like to know if anything turns out to be fragile.


ensime_shared/ensime.py
```diff
@@ -300,3 +362,13 @@ def fun_en_complete_func(self, client, findstart_and_base, base=None):
...
+
+    def watchdog_client_server_connection(self, client):
+        """Run periodically to check and maintain client, server, and cxn health."""
+        if not client.connected and client.connection_attempts < 10:
+            config = client.config
+            server = self.server(config)
+            if not server.isrunning():
+                self.start_server(config)
+            client.connect(server)
+            client.connection_attempts += 1
```
ches: I'm just preserving how the prior implementation behaved so far, but while we're looking at this, it seems excessive to me to check this on every `CursorMove`—`CursorHold` is probably fine, but `CursorMove` a bit much. The server and our connection are not so unstable. Might experiment with removing the call to this from `au_cursor_moved`.

ensime_shared/ensime.py
```diff
@@ -300,3 +362,13 @@ def fun_en_complete_func(self, client, findstart_and_base, base=None):
             # Invoked by vim
             findstart = findstart_and_base
         return client.complete_func(findstart, base)
+
+    def watchdog_client_server_connection(self, client):
```
ktonga: This isn't working (for me), but don't worry, I will take a look as part of #350

ensime_shared/client.py
```diff
@@ -83,38 +86,38 @@ def setup_logger():
...
         self.completion_started = False
 
         self.full_types_enabled = False
         """Whether fully-qualified types are displayed by inspections or not"""
 
-        self.toggle_teardown = True
-        self.connection_attempts = 0
         self.tmp_diff_folder = tempfile.mkdtemp(prefix='ensime-vim-diffs')
 
         # By default, don't connect to server more than once
```
ktonga: Does this comment still make sense here after removing the field?

ensime_shared/ensime.py
ches: Hmm, what about it isn't working?

ensime_shared/client.py
ches: No it doesn't, thanks for pointing that out—I still want to make this stuff clearer. `connection_retries` was a rename from `number_try_connection` because it was clearer to me for how it was used, but it's still confusing that the `connect` method and `queue_poll` are based on counting down `connection_retries`, then what is now `watchdog_client_server_connection` counts *up* `connection_attempts`. I'm guessing that at one time the former was meant to be like a config value for max retries, and the count-up should use that instead of a hardcoded literal, but it perhaps got mixed up over time.

I was pretty much translating the existing code on these up until now, but I'll consolidate them.

ensime_shared/ensime.py
jvican: I think experimentation won't hurt and if it turns out to be unnecessary I am more than welcome to remove it. However, there is a tradeoff here between handlers and responsiveness. If me remove `CursorHold`, `ensime-vim` could turn less responsive and responses may be processed with delays. I hope this is not the case though, I would very much prefer to use only one -- but let's keep in mind that if we're browsing through a file and waiting for some reaction of the plugin, this won't happen until the cursor is on hold for a few secs.

ensime_shared/ensime.py
```diff
@@ -1,24 +1,30 @@
...
-from .launcher import EnsimeLauncher
+from .errors import InvalidJavaPathError
+from .server import Server
 
 
-def execute_with_client(quiet=False,
-                        bootstrap_server=False,
-                        create_client=True):
-    """Decorator that gets a client and performs an operation on it."""
+def execute_with_client():
```
jvican: I'm happy to see these args go away!

ensime_shared/ensime.py
```diff
@@ -80,44 +91,89 @@ def teardown(self):
...
-        if abs_path in self.clients:
-            client = self.clients[abs_path]
-        elif create_client:
-            client = self.create_client(config_path)
-            if client.setup(quiet=quiet, bootstrap_server=bootstrap_server):
-                self.clients[abs_path] = client
-        return client
+        key = os.path.realpath(config_path)
+        if key in self.clients:
+            return self.clients[key]
```
jvican: Do you think not using `return` may improve readability? I usually avoid several `return`s because they break program logic in unexpected ways and they have a big effect on readability in big chunks of code (as readers lose the thread of the logic and can be easily missed). Though redundant, I usually prefer this style:
```scala
client = None
if key in self.clients:
  client = self.clients[key]
else:
  client = self.create_client(config_path)
return client
```
  
The previous code snippet is very common in Java code, where the idiom of "only return once" is quite used. I'm not aware of Python devs using it and I don't want to bikeshed --all this refactoring is amazing--, but I'm just curious what you think about it.

ensime_shared/ensime.py
ktonga: Dunno if it should support this case, but I tried killing the server processes and the client goes to a bad state just logging websocket not connected continuously. I have to restart vim session.
But this is related to my PR where I have to make some changes so recover from a disconnection is possible.

ensime_shared/ensime.py
ktonga: I think ches meant remove just the reconnection logic from `CursorHold`, if `unqueue_and_display` is left there we won't lose any responsiveness.

ches: Yes indeed, the queue processing remains on `CursorMoved`, but checking that server is up and connected is now removed and it seems fine to me.

ensime_shared/ensime.py
ches: Ah okay yeah, I hadn't gone that far but now I have while trying to determine if something else was a bug… the latest commit probably gets you even closer (it does some of the same things you started doing on #350 in fact) but I think that `EnsimeProcess` has some wrong behavior that needs to be fixed in order for `Server` to work right when it has a stale reference to a process. I might leave that for you to pick up in the interest of keeping the scope in check here.

README.md
```diff
@@ -1,7 +1,6 @@
-# Ensime for vim `ensime-vim`
+# `ensime-vim` – ENSIME with Vim!
```
ches: Agree that makes sense, improve the SEO juju 😄  I'll change this.

ensime_shared/ensime.py
ches: Yeah these args and the related `EnsimeClient.setup` method, probably my happiest moments of making these changes 😁  🍾

ensime_shared/client.py
```diff
@@ -85,164 +88,144 @@ def setup_logger():
...
         self.editor = editor
-        self.launcher = launcher
+        self.config = config
 
         self.log = setup_logger()
         self.log.debug('__init__: in')
         self.editor.initialize()
 
         self.ws = None
-        self.ensime = None
```
ktonga: This change breaks `:EnClients`

ensime_shared/ensime.py
```diff
@@ -279,11 +336,15 @@ def au_buf_leave(self, client, filename):
 
     @execute_with_client()
     def au_cursor_hold(self, client, filename):
-        client.on_cursor_hold(filename)
+        """Handler for CursorHold autocommand event."""
+        self.watchdog_client_server_connection(client)
+        client.unqueue_and_display(filename)
+        client.editor.cursorhold()
```
ktonga: This call to editor's `cursorhold` causes the event to be retriggered every second when no activity is happening. I don't think this is the intended behaviour.

ensime_shared/client.py
ches: Indeed, I was aware but hadn't left a todo comment, thanks for logging it here.

`:EnClients` is currently undocumented, I think it should be renamed to `:EnStatus` to reflect that it's really as much about server statuses, and the output formatting enhanced. This could even become a special filetype window for a "console" in our server lifecycle plans, with some summarized key mappings for performing actions like restarting the server on the current line, etc.

I'll just make some minimal fix here so it isn't utterly broken, until we give `EnsimeProcess` some attention in forthcoming work.

ensime_shared/ensime.py
ches: This was moved as-is from `EnsimeClient.on_cursor_hold` so it's not a behavior change in this PR. I don't exactly understand the background of how this achieves "faking a timer" as the link referenced at `Editor.cursohold` describes, but if there's anything wrong with that I'd propose it's brought up in a new issue.

ensime_shared/client.py
jvican: I think that `:EnStatus` would be a better name and it's better to tackle this in another PR.

ensime_shared/ensime.py
jvican: I've been using this PR for the last days. It works well, though I've discovered that vim is slower to respond to keystrokes and, sometimes, events are fired by no reason. For instance, when typing code, the UI gets blocked. I infer this happens because ensime-vim is querying the server and waiting for its response. I don't know if this is a consequence of what @ktonga points out.

jvican: Okay, my issue is that I have a deoplete delay of 450ms that was fired all the time. I removed it. I'll file an issue to keep track of the "blocked UI" issue, as I've experienced it in other ensime requests that take a lot of time to come back.

ktonga: Deoplete while using `omnifunc` does wait for the response in a synchronous way, and more often than I'd like ensime server gets really really slow to respond having to wait several seconds after each keystroke what makes a terrible typing experience, that's why I want to implemente a deoplete source so the integration with autocomplete will be asynchronous avoiding those horrible blocks. But we have to work on ensime-vim integration API first.

ches: Thanks for testing @jvican. Would be curious about reproducible cases of blocking the UI if it's not completions, I think they're *probably* not results of this PR, but still.

In theory these changes could improve responsiveness: there's a subtle change where I moved the `time.sleep` in the polling thread's loop, it was running on every normal iteration and upon inspection this is needless, socket `recv` is efficient blocking I/O that's exactly what we want to happen with no artificial pause for this threading model. I moved the sleep to instead prevent a total busy-loop in the condition that we've lost the connection. This reconnection stuff isn't perfect yet but we'll focus on that soon; I'm not positive that there aren't race conditions (i.e. the `running` and `connected` attributes shared between threads without a queue), but these things haven't really changed here—a deep refactoring of this logic wasn't my intention, yet. I haven't actually been able to observe what I think is a race, the GIL may be solving it. If we're noticeably blocking Vim anywhere though, that's got to be the main thread.

I've been traveling this week so haven't had a chance to tie up the loose ends on this PR, will try this weekend.

ktonga: It's isn't our fault this time, we're only the middleman here. The responsibility for the blocking is shared between Ensime server which from time to time likes taking a nap before responding, and the naive omnicompletion API Vim provides (and we hook into) which is called synchronously from (I guess since it blocks) the UI thread.

ktonga:
Hey @ches do you have any outstanding work on this PR to be pushed? I can rebase it and merge it if you don't have the time, I think it is a huge improvement over the global design that could unblock a bunch of other improvements and I'd like to get it merged sooner rather than later.

Cheers.

ches:
I do have quite a bit more based on top of this that's sitting WIP actually, and I've been using it for awhile too. I'll try to see what's needed to wrap it up cleanly, won't be able to until after Monday but hopefully can find some time by mid-week.

Perhaps for one thing we can simplify on the caveat noted in the PR description by just going ahead and dropping support for the old server bootstrap installation in this branch—we're 8+ months into sbt-ensime handling that, I'm sure just about everyone is using it now.

ktonga:
Totally, let's stop supporting legacy stuff, it is easy enough for people to keep sbt up to date since the recommended plugin installation is add sbt-ensime as a local/per-developer plugin.

I think waiting a few more days for it after waiting for several months is ok :)

ches:
Rebasing this has been a bit tricky after the timer change from #374 and plus we're on server v2 by default now, etc. I have it working but it's shakier than it was before these changes. I think I'll finish sorting it out today though.

ches:
So, I've pushed rebased commits, but the status is still basically as I noted in my last comment: it has some issues that weren't present before. It mostly works, you should be able to run it if you want to try to help me debug, but I'm seeing a few problems:

- You're told twice that the server is starting, and all the messages about server starting, analyzer/indexer ready, etc. now interrupt you requiring a keypress to acknowledge. Not sure why that's happened, I think usually Vim does this when a message is more than one line long. Maybe we've picked up an extra newline somewhere? I don't see that more than one server process is actually getting started, probably just reporting twice as a result of retries when everything is starting up. Might have even been happening before but it wasn't noticeable without the keypress interruptions.
- We sometimes seem to get into a state where the response queue stops being processed: I can use features normally like type inspection, go to definition, completions, and then it seems like trying to go to a definition outside my project (a dependency or stdlib) gets stuck. In debug log it looks like we get a message back from server, but don't dequeue it. Sometimes I can invoke a completion or something and the queue flushes and I see my pending go to def happen.

The interaction of this branch's changes with the timer introduced in #374 is tricky and I've probably just missed reconciling something correctly when rebasing. I'll keep poking at it to try to fix the issues, if anyone tries it now and sees where problems are coming from do let me know!

PR #370, opened 2017-01-13
Work around a startup error on Vim < 7.4

### LINKED ISSUE #369: Error when adding ensime-vim to vim73 on ubuntu lts 12.04
## 2. step by step instructions to trigger the bug 

A. Install vundle on vim73 on ubuntu 12.04 LTS
B. Install ensime-vim as per steps in the instructions
C. Run :PluginInstall and wait for it to complete
D. Quit vim, and re-start vim
E. See this error in vim:

    $ vim
    Error detected while processing function <SNR>14_DependenciesValid:
    line   12:
      File "<string>", line 6, in <module>
    AttributeError: 'module' object has no attribute 'vars'
    Press ENTER or type command to continue

## 3. commit your `.ensime` into the repository (or in a gist)

I don't get that far. This is a problem with ensime-vim, causing an error on start-up of vim.

Here is my .vimrc:

    $ cat ~/.vimrc
    set nocompatible
    filetype off
    
    " set the runtime path to include Vundle and initialize
    set rtp+=~/.vim/bundle/Vundle.vim
    call vundle#begin()
    
    " let Vundle manage Vundle, required
    Plugin 'VundleVim/Vundle.vim'
    
    Plugin 'jewes/Conque-Shell'
    Plugin 'ensime/ensime-vim'
    Plugin 'derekwyatt/vim-scala'
    
    " All of your Plugins must be added before the following line
    call vundle#end()            " required
    filetype plugin indent on    " required

    " My Stuff
    set expandtab
    set hidden
    set ts=4
    set ignorecase
    set sw=4

Disabling ensime/ensime-vim makes the error not print.
Disabling Conque-Shell or vim-scala does not make the error not print.

HEAD of .vim/bundle/ensime-vim is this:

    commit 1b32e85f5c1883af3946cf2bf7a39f60801f47b1
    Author: Ches Martin <ches@whiskeyandgrits.net>
    Date:   Sat Jan 7 00:54:32 2017 +0700
    
    Fail fast if our python dependencies aren't installed (#368)

The python requirements are installed; I ran as indiceted in the docs: `pip install websocket-client sexpdata`

I think the problem is this; "vars" doesn't appear to be a member of the "vim" module:

    " Fail fast if dependencies are missing, we can't do much useful if so.
    " We need to wrap this in a function, see :help script-here
    function! s:DependenciesValid() abort
        python <<PY
    import vim
    try:
        import sexpdata
        import websocket
    
        vim.vars['ensime_deps_valid'] = True
        del sexpdata # Clean up the shared interpreter namespace
        del websocket
    except ImportError:
        vim.vars['ensime_deps_valid'] = False
        pass
    PY
    
        return g:ensime_deps_valid
    endfunction

Reverting this change (this is the latest commit on master) removes the error.

ches:
Closes #369, refs #368 (1b32e85f)

Yeah, I could have left only the `vim.command` version of the code here and had compatibility with old and new, but there are other uses of Vim 7.4+ Python API in our code and this change does not attempt to fix all of them now. This just avoids a particularly poor failure mode during Vim startup.

I think we ought to officially make versions this old unsupported: detect the version and gracefully reject loading the plugin, informing the user. Serious day-to-day work on Scala projects where you'd want IDE features suggests a workstation-class environment IMO, so ensime-vim ought to be free to make assumptions accordingly. ensime-emacs is rather stringent about its requirements. Ubuntu 12.04 (as in #369) goes end-of-life in about 3 months, so it's barely server-class anymore. A possible policy could be to support the last two Vim minor versions (i.e. 8.0 and 7.4). We can discuss this more on a new issue.

### DIFF
--- a/plugin/ensime.vim
+++ b/plugin/ensime.vim
@@ -21,15 +21,28 @@ endif
 function! s:DependenciesValid() abort
     python <<PY
 import vim
+
+# TODO: officially drop Vim < 7.4 support, inform users and don't load plugin
+VIM74 = hasattr(vim, 'vars')
+
 try:
     import sexpdata
     import websocket
 
-    vim.vars['ensime_deps_valid'] = True
+    if VIM74:
+        vim.vars['ensime_deps_valid'] = True
+    else:
+        vim.command('let g:ensime_deps_valid = 1')
+
     del sexpdata # Clean up the shared interpreter namespace
     del websocket
 except ImportError:
-    vim.vars['ensime_deps_valid'] = False
+    if VIM74:
+        vim.vars['ensime_deps_valid'] = False
+    else:
+        vim.command('let g:ensime_deps_valid = 0')
+
+del VIM74
 PY
 
     return g:ensime_deps_valid


ktonga:
Hey @ches why isn't this PR merged yet? Shall we close it in favor of discarding <7.4 support?

ches:
Sorry, I must have missed the approval, I've been preoccupied in the past few weeks.

I'd say let's take this since it does solve a lousy failure mode right now, and #371 remains to track imposing the version restriction.

jvican:
👍

Issue #373, opened 2017-02-02
Use Vim's timer instead of CursorHold event


Kazy:
Vim and Neovim now have a timer functionality to trigger and repeat a function. The related patch: https://github.com/vim/vim/commit/975b5271eed4fa0500c24a8f37be0b1797cb9db7.

While trying to fix an issue related to `CursorHold` with the `EnType` command whose result was immediatly erased from the command bar, I saw the timer patch. I decided to use it in Ensime-Vim instead of using `CursorHold`. The result is available here: https://github.com/Kazy/ensime-vim/commit/04aea153029d3411fa99f2dda1e6aeeb6f3a8bbe. 

It's a quick implementation and it has not been tested intensively. Considering the timer patch is fairly recent, I prefered to run than through an issue first to discuss its inclusion. Let me know what you think.

jvican:
This is awesome @Kazy, thank you for taking a stab at this. Would you mind opening a PR to study the inclusion of this patch into master? I would guess that we should require all our users to bump up their vim versions to 7.4.1578.
  
Do you know if this works with `neovim`? I'm not sure if the function added in `vim` has the same interface as timers in `neovim`.

Kazy:
Yes it's already in Neovim. Since ensime-vim already has a solution working, we could enable the timer functionality conditionally. You can test for the `+timers` feature.

Issue #366, opened 2017-01-05
[Draft] Proposal: Plugin-based Architecture, Extension API


ches:
This should eventually supersede #64, the idea is to refine notes in this issue into a proposal/spec over time. I might move this to a wiki page as it becomes more spec than background story, so we can track history of edits by collaborators, but at the moment the GitHub wiki is disabled on this project.

_I've mixed terms a bit here—should we adopt nomenclature of ensime-vim "extensions" to avoid confusion when we're talking in the context of Vim "plugins"?_

I was thinking about our current architecture as it has been evolving, and it occurred to me that it might be improved by embracing a more plugin-based approach, and almost by construction we would end up with an extensible API that makes it easier for users to do custom scripting against ensime-vim and thus contribute new features more easily (pretty much the essence of #64).

I'm referring primarily to message handlers, functions invoked when particular messages are received from ENSIME. We began with a monolithic client that did just about everything:

- Implemented low-level client/server communication logic.
- Defined the higher-level feature-based function interface for sending calls to the server.
- Implemented all the logic for each specific response message handler.
- Had a sprawling/entangled interface to Vim functionality for UI interaction.
- Even started the server itself (via `EnsimeLauncher`, but the chain of calls originated from the client).

Piece by piece we're breaking this down, decoupling, making testing easier. It's still in progress.

One tool that we've applied crudely to-date is mixins. Look at `debugger.py` and `typecheck.py` and they're just moving food around on the plate: they're still tightly coupled to `EnsimeClient`, directly meddling in its state as `self`, they just offered some code organization into more manageable files. `protocol.py` is the same story but it was additionally solving the problem of alternate versions. It was a reasonable evolutionary solution for the state of the code at the time.

My thinking is that these things can become plugins to the low-level client core, instead of mixins. Rough ideas:

- Plugins are classes that get access to an `EnsimeClient` by constructor injection.
- Plugins are registered with `Ensime`, and when it creates instances of `EnsimeClient` it tells them all the known plugins.
- We provide a function decorator for plugins like `@handles('SymbolInfo')` to declaratively register handlers for received messages.
- `EnsimeClient` dispatches the appropriate messages to each registered handler when received.
- "Everything is a plugin"—all feature-level API of `EnsimeClient` (just about any function that sends or receives a specific ENSIME message) is moved out of it. This will include current "core" functionality but a relatively focused feature area like debugger support might be a good initial use case for hashing out the design.
[...]

PR #374, opened 2017-02-02
Uses Vim's timer instead of CursorHold trigger

Kazy:
Related issue: https://github.com/ensime/ensime-vim/issues/373

### DIFF
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -147,6 +147,18 @@ function! ensime#com_en_clients(args, range) abort
     return s:call_plugin('com_en_clients', [a:args, a:range])
 endfunction
 
+function! ensime#au_cursor_hold(filename) abort
+    return s:call_plugin('au_cursor_hold', [a:filename])
+endfunction
+
+function! ensime#au_cursor_moved(filename) abort
+    return s:call_plugin('au_cursor_moved', [a:filename])
+endfunction
+
+function! ensime#fun_en_tick(timer) abort
+    return s:call_plugin('en_tick', [a:timer])
+endfunction
+
 function! s:call_plugin(method_name, args) abort
     " TODO: support nvim rpc
     if has('nvim')
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -43,14 +43,6 @@ function! ensime#au_buf_leave(filename) abort
     return s:call_plugin('au_buf_leave', [a:filename])
 endfunction
 
-function! ensime#au_cursor_hold(filename) abort
-    return s:call_plugin('au_cursor_hold', [a:filename])
-endfunction
-
-function! ensime#au_cursor_moved(filename) abort
-    return s:call_plugin('au_cursor_moved', [a:filename])
-endfunction
-
 function! ensime#com_en_no_teardown(args, range) abort
     return s:call_plugin('com_en_no_teardown', [a:args, a:range])
 endfunction
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -160,7 +160,7 @@ function! ensime#au_cursor_moved(filename) abort
 endfunction
 
 function! ensime#fun_en_tick(timer) abort
-    return s:call_plugin('en_tick', [a:timer])
+    return s:call_plugin('fun_en_tick', [a:timer])
 endfunction
 
 function! s:call_plugin(method_name, args) abort
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -39,6 +39,10 @@ function! ensime#au_vim_enter(filename) abort
     return s:call_plugin('au_vim_enter', [a:filename])
 endfunction
 
+function! ensime#au_buf_enter(filename) abort
+    return s:call_plugin('au_buf_enter', [a:filename])
+endfunction
+
 function! ensime#au_buf_leave(filename) abort
     return s:call_plugin('au_buf_leave', [a:filename])
 endfunction
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -170,7 +170,6 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
 
                 try:
                     self.ensime = self.launcher.launch()
-                    self.editor.start_timer()
                 except InvalidJavaPathError:
                     self.editor.message('invalid_java')  # TODO: also disable plugin
 
@@ -603,8 +602,8 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
             self.editor.lazy_display_error(filename)
             self.unqueue()
 
-    def refresh(self, filename):
-        """Handler for event CursorHold."""
+    def tick(self, filename):
+        """Try to connect and display messages in queue."""
         if self.connection_attempts < 10:
             # Trick to connect ASAP when
             # plugin is  started without
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -170,6 +170,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
 
                 try:
                     self.ensime = self.launcher.launch()
+                    self.editor.start_timer()
                 except InvalidJavaPathError:
                     self.editor.message('invalid_java')  # TODO: also disable plugin
 
@@ -602,7 +603,7 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
             self.editor.lazy_display_error(filename)
             self.unqueue()
 
-    def on_cursor_hold(self, filename):
+    def refresh(self, filename):
         """Handler for event CursorHold."""
         if self.connection_attempts < 10:
             # Trick to connect ASAP when
@@ -610,12 +611,6 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
             # user interaction (CursorMove)
             self.setup(True, False)
             self.connection_attempts += 1
-        self.unqueue_and_display(filename)
-        self.editor.cursorhold()
-
-    def on_cursor_move(self, filename):
-        """Handler for event CursorMoved."""
-        self.setup(True, False)
         self.unqueue_and_display(filename)
 
     def vim_enter(self, filename):
--- a/ensime_shared/editor.py
+++ b/ensime_shared/editor.py
@@ -334,21 +334,6 @@ class Editor(object):
 
         return ".".join(fqn)
 
-    def cursorhold(self):
-        """Stuff that EnsimeClient does on CursorHold.
-
-        Todo:
-            This is a temporary placeholder for further refactoring.
-
-            IMO EnsimeClient should not have methods like ``on_cursor_hold``,
-            it's the wrong layer of abstraction for handlers of editor events.
-            I'm not sure where these should go yet though -- maybe the Ensime
-            class since it represents "the Vim plugin" as an entity, iterating
-            through its list of clients and calling the logic for each.
-        """
-        self._vim.eval("timer_start({}, 'EnRefreshMessages', {{'repeat': -1}})"
-                       .format(500))
-
     def display_notes(self, notes):
         """Renders "notes" reported by ENSIME, such as typecheck errors."""
 
--- a/ensime_shared/editor.py
+++ b/ensime_shared/editor.py
@@ -200,8 +200,6 @@ class Editor(object):
     # ftplugin/scala/ensime.vim, custom/dotted filetypes & syntax
     def initialize(self):
         """Sets up initial ensime-vim editor settings."""
-        self._vim.options['updatetime'] = 1000
-
         # TODO: This seems wrong, the user setting value is never used anywhere.
         if 'EnErrorStyle' not in self._vim.vars:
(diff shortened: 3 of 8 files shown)


autoload/ensime.vim
```diff
@@ -43,14 +43,6 @@ function! ensime#au_buf_leave(filename) abort
     return s:call_plugin('au_buf_leave', [a:filename])
 endfunction
 
-function! ensime#au_cursor_hold(filename) abort
```
ktonga: I think we shouldn't discard this implementation totally, we should use timers when `+timers` is present and fallback to this implementation otherwise.

Maybe we can encapsulate the scheduling implementation details inside `Editor` class as much as possible.

ensime_shared/client.py
```diff
@@ -602,7 +603,7 @@ def unqueue_and_display(self, filename):
             self.editor.lazy_display_error(filename)
             self.unqueue()
 
-    def on_cursor_hold(self, filename):
+    def refresh(self, filename):
         """Handler for event CursorHold."""
```
ktonga: This comment should be changed accordingly

ensime_shared/editor.py
```diff
@@ -348,11 +346,8 @@ def cursorhold(self):
             class since it represents "the Vim plugin" as an entity, iterating
             through its list of clients and calling the logic for each.
         """
-        # Make sure no plugin overrides this
-        self._vim.options['updatetime'] = 1000
-        # Keys with no effect, just retrigger CursorHold
-        # http://vim.wikia.com/wiki/Timer_to_execute_commands_periodically
-        self._vim.command(r'call feedkeys("f\e")')
+        self._vim.eval("timer_start({}, 'EnRefreshMessages', {{'repeat': -1}})"
+                       .format(500))
```
ktonga: Why the half of the time of before? Have you tried different values and this is the most suitable?
Also why the time as parameter and everything else inline?

ensime_shared/ensime.py
```diff
@@ -278,12 +278,10 @@ def au_buf_leave(self, client, filename):
...
     @execute_with_client()
-    def au_cursor_hold(self, client, filename):
-        client.on_cursor_hold(filename)
-
-    @execute_with_client(quiet=True)
-    def au_cursor_moved(self, client, filename):
-        client.on_cursor_move(filename)
+    def refresh_messages(self, client):
+        filename = client.editor.path()
+        if filename.endswith(".scala"):
```
ktonga: We also support `java` files

rplugin/python/ensime.py
```diff
@@ -172,13 +171,9 @@ def au_vim_leave(self, *args, **kwargs):
...
         super(NeovimEnsime, self).au_buf_leave(*args, **kwargs)
 
-    @neovim.autocmd('CursorHold', **autocmd_params)
-    def au_cursor_hold(self, *args, **kwargs):
-        super(NeovimEnsime, self).au_cursor_hold(*args, **kwargs)
-
-    @neovim.autocmd('CursorMoved', **autocmd_params)
-    def au_cursor_moved(self, *args, **kwargs):
-        super(NeovimEnsime, self).au_cursor_moved(*args, **kwargs)
+    @neovim.function('EnRefreshMessages', sync=True)
```
ktonga: This is only for Neovime, you should also define this function for Vim in `plugin/ensime.vim`

ensime_shared/client.py
```diff
@@ -170,6 +170,7 @@ def lazy_initialize_ensime():
 
                 try:
                     self.ensime = self.launcher.launch()
+                    self.editor.start_timer()
```
ktonga: I can't see this method implemented in `Editor` I think you just kept the old name `cursorhold`. Also I think that this method could use a more descriptive name, something like `start_refresh_timer`

ensime_shared/editor.py
```diff
@@ -348,11 +346,8 @@ def cursorhold(self):
             class since it represents "the Vim plugin" as an entity, iterating
             through its list of clients and calling the logic for each.
         """
-        # Make sure no plugin overrides this
-        self._vim.options['updatetime'] = 1000
-        # Keys with no effect, just retrigger CursorHold
-        # http://vim.wikia.com/wiki/Timer_to_execute_commands_periodically
-        self._vim.command(r'call feedkeys("f\e")')
+        self._vim.eval("timer_start({}, 'EnRefreshMessages', {{'repeat': -1}})"
```
ktonga: You should keep a reference to the timer id returned by `timer_start` so it can be stopped when the plugin is disabled

ches:
I haven't reviewed the changes closely yet, but just wanted to say that if this PR gets to a state that everyone is happy with quickly, then I can rebase #367 to include it. It's my fault that big PR has been open for so long so I'd rather deal with conflicts than ask @Kazy to do so.

Generally I love the idea of being able to use a proper timer instead of the hacky `CursorHold` stuff, but I guess we probably should keep around the old solution for when `+timer` isn't available. I agree with @ktonga that it would be nice to pull the behavior into an abstraction that has the same Python API regardless of what Vim feature is available to back it. That could be a boon for testing too.

ensime_shared/ensime.py
ches: There are a few lines of code in `fun_en_complete_func` that could be factored out into a function for checking "is the buffer one that ensime-vim supports?". It uses Vim's notion of `filetype`, which I think is a better way to go than file extension.

ensime_shared/editor.py
ches: This makes me think that each `Editor` instance shouldn't maintain its own timer as state. There could be a single timer reference as an attribute of `Ensime`. That's kind of the spirit of the todo comment just above this diff.

There can be more than one live instance of `Editor`—currently one is created for each `EnsimeClient` instance, meaning if you edit two ENSIME-enabled projects in one Vim session you'd end up with two timers. That could be changed (in fact #367 changes it, though I didn't think about that much), but I feel leery of having one `Editor` that needs to be treated as a singleton. As an abstraction it might make perfect sense that there _is_ one editor, but practically speaking singletons get ugly. Running with the extensions idea of #366, it'd be nice if third-party code can simply do `editor = Editor(vim)` for UI conveniences, instead of there being some global. It currently has virtually no significant state.

`Ensime` is already effectively a singleton, so it's nice to keep stuff of this nature there as long as it feels like a reasonable design and not overloaded, I think. Thoughts?

ensime_shared/client.py
ches: `tick` might be an apt name for this?

Kazy:
Thanks all for the comments. As explained in the issue, that was more a proof of concept than a proper PR. I'll implement your remarks in the next few days to get this PR in an acceptable state.

ensime_shared/editor.py
Kazy: It's based on nothing, I'll revert back. And the parameter is a left over from the `updatetime` config variable.

ensime_shared/client.py
Kazy: Some uncommited modifications, my bad.

Kazy:
I've put the logic about the periodic function into `Ticker`, responsible for starting the timer and calling tick on a client depending if `+timers` is present or not. Based on @ches comment, the ticker is only present in the main Ensime class, calls the `tick` method on each client with the current filename, and repeats the timer if needed.
Let me know what you think, in particular about the name (if you prefer to revert to simply `refresh_messages`) and about having all that logic in a separate class. 

I've let the cursor hold and cursor move handler for now, so they trigger the `tick` method as well. I wanted to dynamically create the autocmd for those depending on the timer value, but 1) you can't define them after the Neovim plugin class has been created 2) I had issues depending when I called `_vim.eval('timers')`, making it hard to know where to initialize the ticker. I guess I ran into the race condition mentioned in the Ensime class comment.

A solution would be to define those purely in Vim, including starting the timer.

ensime_shared/ensime.py
```diff
@@ -48,13 +50,19 @@ def __init__(self, vim):
...
         self._vim = vim
+        self._ticker = None
         self.clients = {}
 
     @property
     def using_server_v2(self):
         """bool: Whether user has configured the plugin to use ENSIME v2 protocol."""
         return bool(self.get_setting('server_v2', 0))
 
+    def is_buffer_ensime_compatible(self):
```
ktonga: Please move this method to `Editor`, we are trying to reduce the dependency on the Vim object to only the editor module. This is the best for many reasons, but mostly for making the testing easier.

ensime_shared/ensime.py
```diff
@@ -274,22 +291,24 @@ def au_vim_leave(self, client, filename):
...
+    @execute_with_client()
+    def au_buf_leave(self, client, filename):
+        client.buffer_leave(filename)
 
     @execute_with_client()
     def fun_en_complete_func(self, client, findstart_and_base, base=None):
         """Invokable function from vim and neovim to perform completion."""
-        current_filetype = self._vim.eval('&filetype')
-        if current_filetype not in ['scala', 'java']:
+        if not is_buffer_ensime_compatible(self._vim):
```
ktonga: I think this is a good time to achieve what [this TODO](https://github.com/Kazy/ensime-vim/blob/7aaee1440f0c87f94150507c553e599671392d65/ensime_shared/editor.py#L208) says and get rid of this unnecessary check

ensime_shared/util.py
```diff
@@ -41,6 +41,11 @@ def extract_package_name(lines):
                 break
         return package
 
+def is_buffer_ensime_compatible(_vim):
```
ktonga: Please move this method to Editor, we are trying to reduce the dependency on the Vim object to only the editor module. This is the best for many reasons, but mostly for making the testing easier.

ktonga:
Hey @Kazy are you still keen/able to wrap this PR up?

If you rebase it on latest master we maybe could ask the folks in the gitter channel to try the branch for a few days, I can test it on neovim myself.

Cheers.

Kazy:
Yes sure, sorry for the delay. I should be able to do that in the next few days.

ktonga:
@Kazy I think it would be useful if you add a few logging (debug) lines, we could use a bit of info about what's going on, this kind of things tend to be difficult to troubleshot :)

plugin/ensime.vim
```diff
@@ -61,8 +61,10 @@ augroup ensime
     autocmd VimLeave *.java,*.scala call ensime#au_vim_leave(expand("<afile>"))
     autocmd VimEnter *.java,*.scala call ensime#au_vim_enter(expand("<afile>"))
     autocmd BufLeave *.java,*.scala call ensime#au_buf_leave(expand("<afile>"))
-    autocmd CursorHold *.java,*.scala call ensime#au_cursor_hold(expand("<afile>"))
-    autocmd CursorMoved *.java,*.scala call ensime#au_cursor_moved(expand("<afile>"))
+    if !has('timers'):
```
Kazy: I didn't find a way to conditionally define autocmd using Neovim's library. More precisely, `neovim.autocmd` decorator is used when the class (`NeovimEnsime` here) is loaded. You can't call it after the fact to define an autocmd.
A possibility would be to declare the CursorHold and CursorMoved autocmd in `ensime.vim` for Neovim and not in Python.

ktonga: Ok @Kazy, I've been asking to the nvim guys what to do about it ([start of chat](https://gitter.im/neovim/neovim?at=58cf100a7b3f37e75414e44b)) and we should be ok not defining auto commands when in nvim since `timers` have been around for an year now and people already switched to nvim like new stuff that's for sure :)
So let's remove cursorhold and friends from nvim as it is not acceptable to have both mechanisms running at the same time.
What you reckon? I can test it for a while after you change it.

Cheers.

Kazy: Thanks for getting info on that ! It's been implemented. I had to create an autocmd on BufEnter that does nothing to trigger the creation of a client and to start the trigger.

ktonga:
Thanks a lot @Kazy it's looking great. I'll give it a go for a couple of days.

Since it's not a trivial change I could use at least a second pair of eyes :) @ches @jvican anything to add? If you are using anything else other than Neovim it would be nice guys if you could try this branch for a while.

Let's sit on it for a few days and if everything goes fine we are ready to merge it.

Cheers.

ktonga:
@Kazy bad news, for me in neovim the server is never starting but the plugin is activated tho, as in when I trigger a plugin action it tries to send the request but obviously nothing is happening as there is no server to handle it. Nothing on the logs that can give me a clue.

ktonga:
Sorry 'bout the false alarm, it was an unrelated problem.

ktonga:
Hey @Kazy I'm cool with this for Neovim, how was it going for you on Vim? If it went ok we could merge it.

Kazy:
One of my last commit fixed the support for vim. I haven't used it extensively though, just tried a few commands.

ktonga:
Are you keen to fix any bug that could come up? I'm keen to merge it already.

jvican:
@ktonga Let's just merge it. If a bug comes up, we can fix it later down the road. This PR has been for too long in the queue :smile:.

ktonga:
Thank you so much @Kazy for this valuable contribution!

Kazy:
Yes I'm available if bugs come up. Ping me if you see one that might be related !

ches:
Thanks @Kazy, this turned out really nicely.

Issue #372, opened 2017-01-27
ensime-vim doesn't like my path which has spaces and parens in it. 


jsatk:
This appears related to this issue. https://github.com/ensime/ensime-vim/issues/346

I symlink my dotfiles including my `~/.vim` directory. (Maybe I shouldn't symlink that one but that's a story for a different day.) The symlink path is `~/Dropbox (Personal)/dotfiles/vim/`. After spending many hours trying everything under the sun I discovered, thanks to @ktonga's patient help, that it was the path. When I ran `cp -r ~/Dropbox (Personal/dotfiles/vim ~/.vim` and then re-opened vim everything worked lovely. 

After running vim with logging turned on `vim -V9myVim.log` I discovered this suspect part of the log. 

```log
chdir(/Users/jesseatkinson/.vim/bundle/ensime-vim/autoload)
fchdir() to previous dir
line 0: sourcing "/Users/jesseatkinson/.vim/bundle/ensime-vim/autoload/ensime.vim"
Error detected while processing /Users/jesseatkinson/Dropbox (Personal)/dotfiles/vim/bundle/ensime-vim/autoload/ensime.vim:
line    2:
E172: Only one file name allowed: silent! pyfile /Users/jesseatkinson/Dropbox (Personal)/dotfiles/vim/bundle/ensime-vim/autoload/ensime.vim.py
finished sourcing /Users/jesseatkinson/.vim/bundle/ensime-vim/autoload/ensime.vim
continuing in CursorMoved Auto commands for "*.scala"
                                                     Error detected while processing function ensime#au_cursor_moved[1]..<SNR>115_call_plugin:
line   15:
E605: Exception not caught: <type 'exceptions.NameError'>:name 'ensime_plugin' is not defined
Executing CursorHold Auto commands for "*.scala"
autocommand call ensime#au_cursor_hold(expand("<afile>"))
```

Please let me know if I can be of assistance in helping patch this.

ches:
Hi @jsatk, can you try adding an [`fnameescape`][1] to the offending line (where `pyfile` is called), wrapping the `expand`? If that resolves it, a PR would be great.

[1]: http://vimdoc.sourceforge.net/htmldoc/eval.html#fnameescape%28%29

jsatk:
Gladly @ches. I'll do that today or this weekend.

jsatk:
PR here: https://github.com/ensime/ensime-vim/pull/375

PR #375, opened 2017-02-06
Escape path when calling `pyfile`

jsatk:
Addresses: https://github.com/ensime/ensime-vim/issues/372#issuecomment-277273987


jvican:
Thanks for this @jsatk! Let's :ship: it right away since it's a relative short fix :smile:.

PR #381, opened 2017-03-06
Honor selection when asking for type info. Closes #379

### LINKED ISSUE #379: Show type information under selection (not under point)
In `Scala` If there's a line that says:
```
myRepository.getMergeCommitParents(myCommit.id).map(_.map(_.id).toList)
```
The return type `getMergeCommitParents(CommitId)` is a `List[Commit]`
The return type of the whole line is `List[CommitId]`

When I run `EnType` while on `getMergeCommitParents` I get a `List[Commit]` 
There's no way I can run `EnType` on the whole line selection, neither there is a different command that does type information under selection.

You can see what I am referring to in intelliJ by invoking `Alt+ = (Ctrl+ Shift+ P for Mac)`

ktonga:
Fixes #379 by adding a new parameter to `EnType` command which can be used in mappings for applying the command to the visual selection instead of the word under the cursor.

Since there is no way to detect if a selection was active when in `command-mode` and `-range` is only useful line-wise, I had to introduce an optional argument to `:EnType` to know when to operate on the selection instead of on word under cursor. This new arg enables mappings like the following:

```VimL
  nnoremap <buffer> <silent> <Leader>st :EnType<CR>
  xnoremap <buffer> <silent> <Leader>st :EnType selection<CR>
```

Cheers.

### DIFF
--- a/doc/ensime.txt
+++ b/doc/ensime.txt
@@ -174,7 +174,9 @@ available.
                                                                      *:EnType*
 :EnType
 
-    Displays type of the expression under the cursor.
+    Without parameters it displays type of the expression under the cursor.
+    If 'selection' parameter is included it displays the type of the expression
+    enclosed by the last active visual selection.
 
                                                                 *:EnTypeCheck*
 :EnTypeCheck
@@ -404,12 +406,14 @@ autocommands like this in a monolithic `vimrc`: >
 
     autocmd FileType scala,java
           \ nnoremap <buffer> <silent> <LocalLeader>t :EnType<CR> |
+          \ xnoremap <buffer> <silent> <LocalLeader>t :EnType selection<CR> |
           \ nnoremap <buffer> <silent> <LocalLeader>T :EnTypeCheck<CR>
 
 or the equivalent simple mappings in a `$MYVIMRUNTIME/after/ftplugin` file: >
 
     " ~/.vim/after/ftplugin/scala.vim
     nnoremap <buffer> <silent> <LocalLeader>t :EnType<CR>
+    xnoremap <buffer> <silent> <LocalLeader>t :EnType selection<CR>
     nnoremap <buffer> <silent> <LocalLeader>T :EnTypeCheck<CR>
 
 This is a matter of preference. The former may be easier for sharing common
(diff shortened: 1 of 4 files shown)


sherifkandeel:
wow! that was fast... thank you so much @ktonga 

I gave it a shot and it seems like it's working with most cases. However, some of the cases doesn't retrieve the correct type or -weirdly enough- it's retrieving type of different expression.

I think it's either a server issue or an issue from my side (I had some errors/warnings during running `ensimeConfig` and `ensimeConfigProject` since some of the dependencies are on a private maven repo). because selection itself is working fine even on multiple lines.

Unfortunately I tried it on some work code, so I'm not in liberty to share the codebase or examples that had problems, but I'll try to replicate it on dummy project later to check if the problem presists.

But again, Thank you for great and fast work :)

P.S. I didn't know you were going to be that fast, I gave it a shot today but from a different approach, I was going to open a WIP PR but that's when I already found that you opened one :)
You can check my PR against my fork though sherifkandeel/ensime-vim#1

Well, at least I have an idea about the codebase, and I can start contributing :)

ktonga:
@sherifkandeel  Yeah, I saw some weird cases as well, but I'm not doing anything fancy here so maybe it is the server to blame, but will debug it better before making any accusations 😊

You should be careful about multi line selection, I don't think it is working since the end line is not used at all for offset calculation. I left a TODO about that.

Sorry for beating you, wasn't my intention, I got obsessed about solving it as soon as you reported it 😓 I tried to claim the task updating the issue as soon as I started working on it precisely to avoid duplicate efforts. Hope that doesn't discourage you from starting contributing.

Cheers.

sherifkandeel:
@ktonga  Do not worry about that, I never checked the issue progress or anything, I was just having some free time this afternoon/evening and I thought I'd give it a go 😄  

Ah yes, I addressed the multiline selection issue in my PR, I refactored `send_at_point_req` to account for both points and calculated the position difference accordingly. 

I really like the project, it's the exact intersection between python, vim and Scala that I really needed, I think you guys will see more of me here soon xD
 
Cheers

ktonga:
@jvican @ches  Hey guys, I've rebased and updated the docs, I think it's ready to merge but I'll need at least one approval. Could you please have a look at it when you have a chance?

Thanks,
Gaston.

jvican:
Perhaps you can rename the commit title and remove 'WIP'? You can merge as soon as you're done 👍

PR #383, opened 2017-03-14
Fix Windows Java classpath to launch the ensime server

drxcc:
This fixes two problems preventing use of ensime with Vim/GVIM for Windows.

1. Although `java-home` was set correctly, the java executable was not
found since windows uses a `.exe` for the java executable.
2. Windows uses `;` as a classpath separator instead of `:`.

Yea! Now I am running with Ensime in Vim on a Windows machine!

--- 

Although not requiring a code changes, there were a few other gotcha's under a Windows setup that may be useful in a readme:

1.   Since the websocket module uses a binary the both vim and python had to be either 32bit or 64bit versions.

1. running `pip` with the `python -m` prefix was required:

          python -m pip install websocket-client sexpdata

### DIFF
--- a/ensime_shared/launcher.py
+++ b/ensime_shared/launcher.py
@@ -163,6 +163,13 @@ class LaunchStrategy:
         null = open(os.devnull, "r")
         java = os.path.join(self.config['java-home'], 'bin', 'java')
 
+        if(os.name == "nt"):
+            java += ".exe" #the java executable has the .exe extension in windows
+            # windows uses ';' instead of ':' for the classpath sep since ':' is reserved (C:\)
+            # for example  "c:\ensime.jar:c:\scala.jar"  =>  "c:\ensime.jar;c:\scala.jar"
+            # the following ".jar:" hack works because only jar's happen to be on the ensime classpath
+            classpath = classpath.replace(".jar:",".jar;")
+
         if not os.path.exists(java):
             raise InvalidJavaPathError(errno.ENOENT, 'No such file or directory', java)
         elif not os.access(java, os.X_OK):
--- a/ensime_shared/launcher.py
+++ b/ensime_shared/launcher.py
@@ -148,8 +148,10 @@ class LaunchStrategy:
         in a way that is otherwise agnostic to how the strategy installs ENSIME.
 
         Args:
-            classpath (str): Colon-separated classpath string suitable for passing
-                as an argument to ``java -cp``.
+            classpath (list of str): list of paths to jars or directories
+            (Within this function the list is joined with a system dependent
+            path separator to create a single string argument to suitable to 
+            pass to ``java -cp`` as a classpath)
 
         Returns:
             EnsimeProcess: A process handle for the launched server.
@@ -161,14 +163,7 @@ class LaunchStrategy:
         log_path = os.path.join(cache_dir, "server.log")
         log = open(log_path, "w")
         null = open(os.devnull, "r")
-        java = os.path.join(self.config['java-home'], 'bin', 'java')
-
-        if(os.name == "nt"):
-            java += ".exe" #the java executable has the .exe extension in windows
-            # windows uses ';' instead of ':' for the classpath sep since ':' is reserved (C:\)
-            # for example  "c:\ensime.jar:c:\scala.jar"  =>  "c:\ensime.jar;c:\scala.jar"
-            # the following ".jar:" hack works because only jar's happen to be on the ensime classpath
-            classpath = classpath.replace(".jar:",".jar;")
+        java = os.path.join(self.config['java-home'], 'bin', ('java','java.exe')[os.name == 'nt'] )
 
         if not os.path.exists(java):
             raise InvalidJavaPathError(errno.ENOENT, 'No such file or directory', java)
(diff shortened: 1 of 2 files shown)


ensime_shared/launcher.py
```diff
@@ -163,6 +163,13 @@ def _start_process(self, classpath):
         null = open(os.devnull, "r")
         java = os.path.join(self.config['java-home'], 'bin', 'java')
 
+        if(os.name == "nt"):
+            java += ".exe" #the java executable has the .exe extension in windows
```
ktonga: Not too happy with the '.exe' hack but I'm kinda ok with it since we manipulate the java binary path.
An idea that occurred to me is that we can only check that java-home exists (it is very unlikely that it exists but it doesn't contain the java bin) and then we can use same java bin path for all OSs, I understand that you can execute a binary in windows without specifying the '.exe' extension. Is that right?

ensime_shared/launcher.py
```diff
@@ -163,6 +163,13 @@ def _start_process(self, classpath):
...
+            # windows uses ';' instead of ':' for the classpath sep since ':' is reserved (C:\)
+            # for example  "c:\ensime.jar:c:\scala.jar"  =>  "c:\ensime.jar;c:\scala.jar"
+            # the following ".jar:" hack works because only jar's happen to be on the ensime classpath
+            classpath = classpath.replace(".jar:",".jar;")
```
ktonga: I think the right place to fix that is in the sbt plugin, in vim we should use the setting as it is. What do you think @fommil ?

I'm really surprised that this is the first time that supporting windows comes up.

fommil:
The sbt plugin gives a list for the classpath, you must be doing something when parsing it. Definitely works on Windows (my day job demands it...)

ktonga:
@fommil you're right, I should have checked before bothering you with stupid mentions. Sorry.

It was implemented as part of bothers-sam-at-work milestone 😂

fommil:
:-D no more Windows for me after tomorrow... woohoo!

Which probably means features instead of lots of hacks.

ensime_shared/launcher.py
ktonga: @drxcc well, I had a better look at where `classpath` is formatted from a list and it is done in multiple places, one per each launcher strategy, like in [this line](https://github.com/drxcc/ensime-vim/blob/f6abb669269646b097fb4765b3bc55d3de11e768/ensime_shared/launcher.py#L234)

I think the right thing to do here is to carry `classpath` as a list until it really needs to be formatted as string, that is right here in `_start_process` where you are doing it, but manipulating it as string seems a bit dirty :), so after changing it to be a list you can , you can pick the proper separator.

What you think? do you want to do the small refactor?

Please note that there is a tests for that which should be update accordingly. You can run the tests with `make test`

drxcc: Ok, I pushed the changes (see 1a86e7128f01afa6adc25578223941cb184c2456) to make the classpath a list and only make it into a string at the argument location. 

I was also able to successfully run the `make test` after changing the classpath there to lists also.   However, in order to run the tests on windows, I had to change the Makefile but kept this [in a different branch](/drxcc/ensime-vim/windows-test).

ensime_shared/launcher.py
drxcc: I didn't like the `.exe` hack at first either, but now after some research and trying out the `java-home` directory existence test I think the `.exe` extension actually is the way to go.  Here is why:

1.  The `.exe`  addition is actually more robust on a windows system then leaving it off.  To make the auto `.exe` extention work, the users `PATHEXT` environment variable has to be properly setup with a `.exe` listing.
2. The `.exe` is required if we still want to test if the file is set as executable for the current user
3. The `java-home` directly only checks if that directory is availble, while it is unlikely the files below would be there it is not necessarily true that since the java_home path is often symlink'ed.

drxcc:
Sure, after the ensime-vim/master is working in Windows, I would be glad to add some notes for windows users on ensime/ensime.github.io

ensime_shared/launcher.py
```diff
@@ -161,15 +163,15 @@ def _start_process(self, classpath):
         log_path = os.path.join(cache_dir, "server.log")
         log = open(log_path, "w")
         null = open(os.devnull, "r")
-        java = os.path.join(self.config['java-home'], 'bin', 'java')
+        java = os.path.join(self.config['java-home'], 'bin', ('java','java.exe')[os.name == 'nt'] )
```
ktonga: I would prefer the more idiomatic:

```python
'java.exe' if os.name == 'nt' else 'java'
```

ensime_shared/launcher.py
```diff
@@ -161,15 +163,15 @@ def _start_process(self, classpath):
...
 
         if not os.path.exists(java):
             raise InvalidJavaPathError(errno.ENOENT, 'No such file or directory', java)
         elif not os.access(java, os.X_OK):
             raise InvalidJavaPathError(errno.EACCES, 'Permission denied', java)
 
         args = (
-            [java, "-cp", classpath] +
+            [java, "-cp", (':',';')[os.name == 'nt'].join(classpath)] +
```
ktonga: Ditto if/else

drxcc:
Thanks, @ktonga.  Ok, I made the simple ternary change.   `make test` works with the newest 940379f  on Windows.

 (If anyone else heads down this route, I was using 64bit Vim8.0 and Gvim8.0 with Windows8 64bit.)

PR #398, opened 2017-07-05
Broken Sexp2dict

VC1995:
As per my conversation with @ktonga 
There was an issue in `ProjectConfig`'s parse method's `sexp2dict` method where incase of list of lists in the loaded sexp, only the first element of the list was considered.


ches:
Always suspected that was too easy 😄  Thanks!

PR #394, opened 2017-04-19
Fix #393

nickdiego:
After f6abb669269646, classpath read from
bootstrap classpath file must be splitted into
a list before to be passed to _start_process.

Signed-off-by: Nick Diego Yamane <nick.diego@gmail.com>

PR #402, opened 2017-08-06
Implement `EnUsages` command which searches for symbol usages

ktonga:


ches:
Woot, looking forward to trying this out!

Separately, still gotta figure out if whatever made the CI build stop working and then come back but not actually run anything can be remediated… https://github.com/ensime/ensime-vim/pull/394#issuecomment-295308150

ensime_shared/protocol.py
```diff
@@ -94,6 +97,12 @@ def handle_type_inspect(self, call_id, payload):
     def show_type(self, call_id, payload):
         raise NotImplementedError()
 
+    def handle_source_positions(self, call_id, payload):
+        raise NotImplementedError()
```
jvican: Should we open an issue so that contributors fill this in at some point of the future?

jvican:
@ches Please, feel free to merge when you approve it :smile:.

ensime_shared/protocol.py
ktonga: Sorry, I'm not following. This is the ugly way for defining an abstract method in python, so it doesn't need to be filled here.

Anyway, after @ches finishes his big refactor I'll rework the handlers, I think resolving them for the server version using inheritance is a terrible design and I want to change it. Also they should depend on protocol version as opposed to on server version.

ktonga:
Thanks @jvican it still has some small annoyances on the server side, nothing to do on vim's side tho, as soon as the results gets better on the server we will see it on vim without effort. For instance I think it still doesn't work on methods.

ensime_shared/protocol.py
jvican: Oh I see, was not familiar with this. Thanks for elaborating.

ensime_shared/editor.py
```diff
@@ -278,8 +278,8 @@ def to_quickfix_item(self, file_name, line_number, message, tpe):
                 "text": message,
                 "type": tpe}
 
-    def write_quickfix_list(self, qflist):
-        self._vim.command('call setqflist({!s})'.format(qflist))
+    def write_quickfix_list(self, qflist, title):
+        self._vim.command("call setqflist({!s}, 'r', 'Ensime - {}')".format(qflist, title))
```
ches: I believe this signature of `setqflist` is an incompatibility between Vim and Neovim. In Vim 8 the third argument is documented as being a dictionary and has this example:

```vim
:call setqflist([], 'r', {'title': 'My search'})
```

It looks like Vim 7 [didn't support it at all][1] and I think giving up Vim 7 support might be a bit premature now, Neovim difference aside.

[1]: http://vimdoc.sourceforge.net/htmldoc/eval.html#setqflist()

ensime_shared/protocol.py
```diff
@@ -247,3 +264,28 @@ def show_type(self, call_id, payload):
...
+        qfList = []
+        for p in positions:
+            position = p["position"]
+            preview = str(p["preview"]) if "preview" in p else "<no preview>"
+            item = self.editor.to_quickfix_item(str(position["file"]),
+                                                position["line"],
+                                                preview,
+                                                "info")
+            qfList.append(item)
+        qfSorted = sorted(qfList, key=itemgetter('filename', 'lnum'))
```
ches: It's coming from other instances elsewhere I know, but we're mixing style of naming conventions with `qfList` and `qfSorted`.

plugin/ensime.vim
```diff
@@ -73,6 +73,7 @@ command! -nargs=* -range EnInstall call ensime#com_en_install([<f-args>], '')
 command! -nargs=* -range EnNoTeardown call ensime#com_en_no_teardown([<f-args>], '')
 command! -nargs=* -range EnTypeCheck call ensime#com_en_type_check([<f-args>], '')
 command! -nargs=* -range EnType call ensime#com_en_type([<f-args>], '')
+command! -nargs=* -range EnUsages call ensime#com_en_type([<f-args>], '')
```
ches: Should be `com_en_usages` 😊

ensime_shared/client.py
```diff
@@ -421,6 +429,15 @@ def doc_uri(self, args, range=None):
         self.log.debug('doc_uri: in')
         self.send_at_position("DocUri", False, "point")
 
+    def usages(self, args, range=None):
```
ches: Oh yeah, I think it was my complaint about this that prompted creation of #244. Don't recall if there's a blocker to stop proliferating the pattern, if not changing all existing ones yet.

ensime_shared/client.py
```diff
@@ -249,7 +249,7 @@ def send_at_position(self, what, useSelection, where="range"):
         self.log.debug('send_at_position: in')
         b, e = self.editor.selection_pos() if useSelection else self.editor.word_under_cursor_pos()
         self.log.debug('useSelection: {}, beg: {}, end: {}'.format(useSelection, b, e))
-        self.send_at_point_req(what, self.editor.path(), b[0], b[1], e[0], e[1], where)
+        self.send_at_range_req(what, self.editor.path(), b[0], b[1], e[0], e[1], where)
```
ktonga: I know, right? the current state is far from ideal, but I tried to reduce the changes not related to the new feature to the minimum. I did take a look at the motivation for such a messy set of functions, and the thing is, that they are like this to hide some inconsistencies from the server's protocol, that arg is used to name the property in the message, some times is called 'range' and some times 'point' but it always carries the 'from' and 'to' that's why I renamed and called `...at_point...` the proper one that actually sends the point.
I'd prefer to leave it like this by now, the sensible refactor to do here would be to introduce a model (with the case classe python equivalent) to represent the server messages instead of cryptic functions which create dics on the fly

ensime_shared/client.py
ktonga: Ditto unrelated changes changes.

I'd prefer to align to the current pattern and change them all later on a tech-debt PR

ensime_shared/editor.py
ktonga: Oh my, really good catch here, dropping Vim 7 support wasn't my intention at all :smile:. I just naively followed nvim docs for the function trusting on them being compatible across vim versions/flavors.

- [ ] Will rework this to make sure that ir runs everywhere (aka add a bunch of `if`s :disappointed:)

ensime_shared/protocol.py
ktonga: WDYM? there's only another instance when `qfList` is used, and introduced `qfSorted` to hold the sorted copy of `qfList`. How you reckon they should be called?

plugin/ensime.vim
ktonga: Sorry, I might have killed a kitten with this ~copy&paste~ yank&put. Will fix it.

- [ ] Fix copipasta

ensime_shared/protocol.py
ktonga: Oh I gotcha! you mean camelCase vs snake_case.

I can look around and turn all names into snakes, maybe there's even a linter setting to ensure that. But again, would prefer to do it on a separate PR since this was not introduced in this PR

ensime_shared/client.py
ches: I understand trying to keep changes minimal, but if this is now the only call site of the renamed `send_at_range_req`, could we drop its `where` parameter? I.e. branch in `send_at_position` if it was called with `where='range'` vs `'point'`. Does it sometimes need to be called with `where='point'` to send a message with `'point': {'from': …, 'to': …}`?

My concern with it is, previously you had one method `send_at_point_req` which could be called with either "point" or "range"… okay, that's a little bit weird to understand, but now you have two methods which *seem* to helpfully separate point vs. range yet actually you still sometimes might need to call the range one with a "point" argument! (Do you?)

I had started modeling protocol messages as `namedtuple`s in my separate client project but the protocol was changing more than I was willing to keep up with… if that's slowed down maybe we could start doing that in ensime-vim in the short term (not in this PR of course).

ensime_shared/client.py
ches: I don't consider things "unrelated changes" though when we're talking about newly-added code. If it's impossible to define this without the unused `args` and `range` params without changing other things to make it work, then okay that's fine. But if it is possible already (that's my question, I didn't try yet), what's the justification to blindly keep adding new code this way?

ensime_shared/editor.py
ches: I unfortunately encountered exactly this API diversion once in the past 😞

ensime_shared/protocol.py
ches: Yep I meant the the camelCase. Sure no need to change other instances outside this PR now, but this case is new right?

ensime_shared/client.py
ktonga: Turns out it is possible and really easy to do. Done!

ensime_shared/editor.py
ktonga: Fixed. I've tested it on both, vim and nvim, and it works.

ensime_shared/protocol.py
ktonga: Done

ensime_shared/client.py
ktonga: Ok, so the thing is that at_range was only used once within send_at_position so I got rid of it and inlined the code in the later which is indeed called with 'point' a few times and in one case it uses the default value:

```
ensime_shared/client.py|357 col 14| self.send_at_position("Type", useSelection)
ensime_shared/client.py|436 col 14| self.send_at_position("DocUri", False, "point")
ensime_shared/client.py|451 col 14| self.send_at_position("DocUri", False, "point")
```

I've added some docs to the method to clarify the use of the parameters, should be a bit clearer now, still not ideal and I would like to improve it but any further refactor is out of the scope of this PR IMHO.

So now we only have two separate send_at... methods and they have clear different responsibilities, one sends `from` and `to` based on selection or word under cursor some times calling the field 'range' and other 'point, the other method simply sends the cursor position coordinates.

ches: Nice, I think that narrowed responsibility of `send_at_point` is an improvement until we get more type-ful requests. Thanks!

ches:
Oh, updating vimdoc would be a nice a finishing touch if you have the chance ☺️

ktonga:
@ches oh you're right, totally forgot about the docs. Done.

Thank you very much for taking the time to review it guys.

PR #411, opened 2018-04-06
Python 3 support

Ch00k:
This makes the code and tests work on both Python 2 and 3.
Besides fixing tests I also ran `2to3` on all Python files, but what the tool did not suggest any meaningful changes.
The test coverage is quite low, so I guess this can only be tested well in the field :)


ktonga:
Awesome @Ch00k!! thanks a lot for taking the time. I'll take a look later and try to use the branch for a while as the litmus test :)

Ch00k:
@ktonga actually I was too quick to announce this. It still has problems that need to be solved. I'll make some more changes to the branch later today. Sorry for confusion .

Ch00k:
Ok, so I'm at the point where I am getting this error while opening a scala file in vim with Python 3:
```
Error detected while processing function ensime#au_buf_enter[1]..<SNR>83_call_plugin:
line   15:
E605: Exception not caught: <class 'NameError'>:name 'ensime_plugin' is not defined
```
The exception is thrown by [this](https://github.com/Ch00k/ensime-vim/blob/6c1c65ff75954dba0978fb3d9bac1d6da0ea40ca/autoload/ensime.vim#L189) line. If I understand correctly, the `ensim_python` should become defined [here](https://github.com/Ch00k/ensime-vim/blob/6c1c65ff75954dba0978fb3d9bac1d6da0ea40ca/autoload/ensime.vim.py#L24), but somehow it does not. Look like vim does not load this Python file at all? I tried to insert some debugging code in the very beginning of `ensime.vim.py` file:
```
with open('/tmp/foo', 'w') as f:
    f.write('OK')
```
With Python 3 `/tmp/foo` does not get created.
With Python 2 though, everything works perfectly.

ktonga:
I tried it in Neovim and it is not working either.

First I had to rename `rplugin/python/` to `rplugin/python3/` so Python3 host is used when running `UpdateRemotePlugins`, but after sorting that out another error occurred, it can't find `ensime_shared` module, I guess it's b/c the folder is at the root instead of being under `rplugin/python3/`

I would recommend let's forget about python2 and focus only on python3 support, it's pretty realistic to think that everyone using vim as an IDE has support for py3 plugins.

Deoplete is a good example of python3 vim/neovim plugin https://github.com/Shougo/deoplete.nvim

Ch00k:
Finally figured out what was wrong: should call `py3file` instead of `pyfile`. Since the command had `!silent`, it was hiding the error :(

Ch00k:
@ktonga I think I fixed it for neovim as well. Could you please try on your end again?

ktonga:
@Ch00k I'm using it and it seems to be working nicely. I'll keep using the branch for a while, little details use to be found when you use things in anger :)

Ch00k:
@ktonga Any more feedback on that? I have another PR that I'd like to suggest when this one gets merged: https://github.com/Ch00k/ensime-vim/pull/1

ktonga:
@Ch00k I have been using it for a while now in neovim and didn't find any problem. Can you say you'll be able to keep responsive to bug reports if I merge it and people start having problems?

Ch00k:
@ktonga I'll try me best.

PR #412, opened 2018-04-21
Python dependencies as submodules

Ch00k:
This adds all Python dependencies as git submodules. They will all be cloned upon plugin installation, and the user will no longer need to take any manual steps, i.e. `pip install`.

### DIFF
--- a/autoload/ensime.vim
+++ b/autoload/ensime.vim
@@ -1,5 +1,6 @@
 if !has('nvim')
-    execute 'silent! py3file' fnameescape(expand('<sfile>:p').'.py')
+    execute 'py3file' fnameescape(expand('<sfile>:p').'.py')
+    "execute 'silent! py3file' fnameescape(expand('<sfile>:p').'.py')
     execute ':redraw!'
 endif
 
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -11,7 +11,6 @@ import time
 from subprocess import PIPE, Popen
 from threading import Thread
 
-import websocket
 
 from .config import feedback, gconfig, LOG_FORMAT
 from .debugger import DebuggerClient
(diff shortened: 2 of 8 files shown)


autoload/ensime.vim
```diff
@@ -1,5 +1,6 @@
 if !has('nvim')
-    execute 'silent! py3file' fnameescape(expand('<sfile>:p').'.py')
+    execute 'py3file' fnameescape(expand('<sfile>:p').'.py')
```
Ch00k: I think it's better not to hide the output here. Makes debugging a bit easier.

Ch00k:
Any feedback on this?

ktonga:
I don't think it's necessary. You have to pip install neovim module anyway, so I don't see how adding two more names to the list can be problematic.

ktonga:
I'll give it a spin today. Could you please raise a PR to update [the installation instructions](http://ensime.github.io/editors/vim/install/) in the meantime? Please also include the recent change py2 -> py3

Thanks!

autoload/ensime.vim
ktonga: Ok. Remove commented line then.

ensime_shared/client.py
```diff
@@ -26,6 +25,12 @@
 else:
     from Queue import Queue
 
+six_path = os.path.join(os.path.dirname(__file__), 'six')
+sys.path.insert(0, six_path)
+websocket_path = os.path.join(os.path.dirname(__file__), 'websocket-client')
+sys.path.insert(0, websocket_path)
+import websocket
```
ches: I think it'd be nice to hide away all the `sys.path` hackery in one place at early initialization, so you can keep normal-looking imports in the commonly-edited files and probably avoid linter complaints, etc. The `__init__.py` in `ensime_shared` might work, or it might have to go to `autoload/ensime.vim.py` and `rplugin/python3/ensime.py` where there is already path hackery. There is some duplication in that case, maybe there's a way to eliminate it but I didn't find a good one yet.

ches:
Also, why do we need six at this time?

PR #413, opened 2018-06-06
use asyncio for websocket communication

svalaskevicius:
the timeouts for polling add up to enormous waiting times, especially on pure FP files where ensime sends a lot of ScalaNote messages for libraries it parses.

**this pr is not finished**. please excuse the commented blocks left here etc, this is rather an attempt (successful it seems!) to make it faster by using asyncio instead of periodic polls..

any comments are welcome, before I try clean it up and make it backwards compatible (as asyncio only exists from python 3.4 apparently, and python 3.5 added some syntax used here)..

also.. while this pr increased my python-fu considerably, it's still rather basic.. **any help from ppl who want to join would be appreciated greatly!**

### DIFF
--- a/ensime_shared/client.py
+++ b/ensime_shared/client.py
@@ -11,7 +11,6 @@ import time
 from subprocess import PIPE, Popen
 from threading import Thread
 
-import websocket
 
 from .config import feedback, gconfig, LOG_FORMAT
 from .debugger import DebuggerClient
@@ -21,10 +20,15 @@ from .typecheck import TypecheckHandler
 from .util import catch, Pretty, Util
 
 # Queue depends on python version
-if sys.version_info > (3, 0):
-    from queue import Queue
-else:
-    from Queue import Queue
+#if sys.version_info >= (3, 5):
+import asyncio
+import websockets
+#elif sys.version_info > (3, 0):
+#    from queue import Queue
+#    import websocket
+#else:
+#    from Queue import Queue
+#    import websocket
 
 
 class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
@@ -101,7 +105,11 @@ class EnsimeClient(TypecheckHandler, DebuggerClient, ProtocolHandler):
         self.refactorings = {}
 
         # Queue for messages received from the ensime server.
-        self.queue = Queue()
+#        if sys.version_info >= (3, 5):
+        self.ws_loop = None
+        self.queue = None
+#        else:
+#            self.queue = Queue()
         self.suggestions = None
         self.completion_timeout = 10  # seconds
         self.completion_started = False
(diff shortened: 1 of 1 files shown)


ensime_shared/client.py
```diff
@@ -593,38 +619,56 @@ def type_check(self, filename):
...
-                result = self.queue.get(False)
-                self.log.debug('unqueue: result received\n%s', result)
-                if result and result != "nil":
-                    wait = None
-                    # Restart timeout
-                    start, now = time.time(), time.time()
+        with catch(Exception, self.log.error):
+            if timeout != 0 or not self.queue.empty():
+                if timeout == 0:
+                    timeout = 0.01
```
svalaskevicius: should probably reintroduce should_wait param... although... maybe another function might be better for that?

ensime_shared/client.py
```diff
@@ -219,14 +243,16 @@ def disable_completely(e):
...
-                self.ws = websocket.create_connection(self.ensime_server, **options)
-            if self.ws:
+                async def connect():
+                    options = {"subprotocols": ["jerky"]} if server_v2 else {}
+                    self.log.debug("About to connect to %s with options %s",
+                                self.ensime_server, options)
+                    self.ws = await websockets.client.connect(self.ensime_server, **options)
+                    True
+                gotws = asyncio.run_coroutine_threadsafe(connect(), self.ws_loop).result(10)
+            if gotws:
```
svalaskevicius: is this `gotws` needed at all? if future fails it will throw exception, so should be ok w/o it and the following line shifted right into the catch block

ensime_shared/client.py
```diff
@@ -101,7 +105,11 @@ def setup_logger():
         self.refactorings = {}
 
         # Queue for messages received from the ensime server.
-        self.queue = Queue()
+#        if sys.version_info >= (3, 5):
+        self.ws_loop = None
```
svalaskevicius: should probably move to a new class, that handles the other thread and exposes thread safe methods

svalaskevicius: added benefit - diff classes can be used for diff python versions

ches:
Hi @svalaskevicius, thanks for working on this! I made an attempt at converting to asyncio awhile back (partly in a separate project that I've never gotten as far as PR for, I was going to make a pip-installable Python ENSIME client that ensime-vim + the Sublime plugin could use as a dependency potentially), but I wrestled with trying to keep Python 2 support through backport implementations.

It's more than a year later now and I think it's probably beyond the point of being worth trying to keep Python 2 compat now, so I'd be happy to see this move forward. The Trollius project is basically dead, and I think it's the norm now for Vim to build with Python 3 support by default in most desktop-oriented distributions. I know it is the case for Homebrew's `macvim` package on OS X, I need to specify a `with-python@2` build option now to keep plugins like ensime-vim working.

By the way, I was using the same websockets library that you chose as well.

I'll try to give this some review soon if you're ready for it, sorry if it takes some time, the maintainers haven't been able to spare a lot of attention to the project lately.

svalaskevicius:
Cool! Agreed re Python 2. What about python <3.5? Do you know if it's used widely or can we use this async syntax I've used here?

As I mentioned, any review comments are welcome even if it's not finished - would help to guide the design thinking (I've added a few of my own :))

svalaskevicius:
Also, re time investment - I'm not sure how much time I'll have either, so no pressure. The basic idea for me was to get it to a usable state, but I'm still to try the LSP approach :)

svalaskevicius:
hmm, removed ticker too in https://github.com/svalaskevicius/ensime-vim/commit/8387c47ecdbed600ee5014f63e31b014df13cc9e - seems like it's not really required, and this way ensime feels much faster. as fast is it probably can be..

is ticker useful for anything else than lazy_display_errors? if that switched to quickfix list (separate pr discussion, see #415), it could be removed altogether... it seems - still testing/experimenting :)

Issue #408, opened 2017-10-05
Type check errors in quickfix window


mlopes:
When running :EnTypeCheck, it highlights the errors and then shows them when passing with the cursor over the code causing the error. This error is being shown in the command area, which means we get a message "Press ENTER or type command to continue" unless we set command height to at least 4 (the size of the output of a EnTypeCheck error).
I'm wondering if using vim's quickfix window could be a better option, allowing the user to quickly navigate through typecheck errors and not get stuck on the error line every time they pass over it?

ches:
Hi Marco,

ensime-vim has built-in Syntastic integration support that essentially
provides quickfix lists for this—I think most users use this, or perhaps
wire up integrations with alternatives like Unite.

That said though, I’d be glad to consider PRs for improving how this works
out-of-the-box with no other third-party plugin (or shipping integrations
in addition to Syntastic with ensime-vim). I believe we might have another
enhancement ticket along these lines already, I’m on the go at the moment
so I haven’t searched.

>

mlopes:
I’m using ale for syntax checking. Not sure if there will be an easy way to integrate. I’ll have to look into what are my options.

Thanks for your reply.

mlopes:
Ok, I might have a go at making this work. It seems like a checker can be added to ale's registry:

`ale#fix#registry#Add(name, func, filetypes, desc)      *ale#fix#registry#Add()*`

Now this is where I get a bit confused. In `ensime-vim`, a checker is registered with syntastic:

https://github.com/ensime/ensime-vim/blob/dbf5976a5cf39829fd78e1285f68ab0eefef5a0b/syntax_checkers/scala/ensime.vim#L28

but it sets `'exec': '/bin/true` and I can't really see how `syntastic` gets the data from `ensime-vim` when for example a type check fails.

BonarBeavis:
I am trying to integrate ensime with neomake. Neomake does support callback function for a maker, but I do not know which function to call. Any hint?
I looks like the syntastic integration gets it data from `b:ensime_notes` which exists but is always empty.

MaxWorgan:
@mlopes did you get anywhere with this - I'm trying to achieve the same thing.

mlopes:
Hi, 

No, I know very little about vim plugin development, never really got to the bottom of how this integrates with syntastic. It doesn’t look like it should be to hard though.

mlopes:
[...]

PR #415, opened 2018-06-06
avoid the annoying hit-return issue on messages

svalaskevicius:
a quickfix to avoid hit-return messages in vim (neovim) when echoing messages of correct length.

disables showcmd and ruler as they are being checked here: https://github.com/neovim/neovim/blob/7795829767818a3c6cbeed7957a1ad5c03d48d41/src/nvim/option.c#L5284 
once echoed, restores the state

### DIFF
--- a/ensime_shared/editor.py
+++ b/ensime_shared/editor.py
@@ -321,6 +321,9 @@ class Editor(object):
         cmd = 'echo "{}"'.format(message.replace('"', '\\"'))
         if silent:
             cmd = 'silent ' + cmd
+        cmd = "let _ensime_showcmd=&showcmd | set noshowcmd | let _ensime_ruler=&ruler | set noruler | " + \
+		cmd + \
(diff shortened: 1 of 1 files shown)


svalaskevicius:
alternative idea is to use a preview window.. and show full message instead of trying to cram it in the status line :) something like this: https://github.com/pushcx/ale/pull/1/files just maybe adjusting the height to the lines required too

mlopes:
Merging this PR would mitigate the issue I described on issue #408 where the pluggin becomes completely broken (the user is asked to repeateadly press enter preventing them from editing the offending line) if the user doesn't use syntastic.

svalaskevicius:
actually, using quickfix feels quite good, more usable than the ticker based scroller - experimenting with it in my branch - https://github.com/svalaskevicius/ensime-vim/commit/c09fbed325d19f8cba8082412fb70e2cd3ac7960

will continue testing it for time being (as that branch is quite hacky/experimental still), could make a separate PR if anyone interested?

